PR #3, opened 2015-09-02
Ticket #30818 - Fix for the wrong base pipeline configuration

robblau:
The commands that are being run are being launched from Desktop, which would have a TANK_CURRENT_PC env var set to the site configuration.  Since we preserve that value for subprocesses (which is usually the behavior we want), the DCCs being launched were trying to run in the project environment and getting an error due to the conflict.

This change updates the logic for launching commands from the webserver to make sure that subprocesses run with a clean environment.

### DIFF
--- a/python/tk_framework_desktopserver/command.py
+++ b/python/tk_framework_desktopserver/command.py
@@ -52,7 +52,13 @@ class Command(object):
             else:
                 startupinfo = None
 
-            # Clean out Toolkit specific environment variables to launch the process cleanly
+            # The commands that are being run are probably being launched from Desktop, which would
+            # have a TANK_CURRENT_PC environment variable set to the site configuration. Since we
+            # preserve that value for subprocesses (which is usually the behavior we want), the DCCs
+            # being launched would try to run in the project environment and would get an error due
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/command.py
```diff
@@ -50,10 +51,19 @@ def call_cmd(args):
                 startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
             else:
                 startupinfo = None
+
+            # Clean out Toolkit specific environment variables to launch the process cleanly
```
manneohrstrom: It would be great if we could add the pull request description to the comment - i think it would help future readers of the code:

```
The commands that are being run are being launched from Desktop, which would have a TANK_CURRENT_PC env var set to the site configuration. Since we preserve that value for subprocesses (which is usually the behavior we want), the DCCs being launched were trying to run in the project environment and getting an error due to the conflict.

This change updates the logic for launching commands from the webserver to make sure that subprocesses run with a clean environment.
```

manneohrstrom:
Looks good! One minor suggestion.

PR #4, opened 2015-09-16
For #31081, cleanly shutdown websockets

jfboismenu-adsk:
- Correctly waits for the server thread to end on shutdown.
- Forwards the Twisted logs to the Python logs.

### DIFF
--- a/python/tk_framework_desktopserver/server.py
+++ b/python/tk_framework_desktopserver/server.py
@@ -96,17 +96,12 @@ class Server:
         self._reactor_thread = threading.Thread(target=start)
         self._reactor_thread.start()
 
-    def start(self, start_reactor=False):
+    def start(self):
         """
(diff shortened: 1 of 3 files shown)


python/tk_framework_desktopserver/server.py
```diff
@@ -98,8 +108,15 @@ def start(self, start_reactor=False):
...
+        """
+        return self._reactor_thread.isAlive()
+
+    def tear_down(self):
         """
-        Stops the serverreactor.
+        Tears down Twisted.
         """
         reactor.callFromThread(reactor.stop)
+        self._reactor_thread.join()
```
robblau: When is start_reactor false in the args to start?

Given the current code if it is false, then I think self._reactor_thread won't be defined when tear_down is called and this will error.

PR #5, opened 2015-09-29
For #30860, make websocket errors more obvious

jfboismenu-adsk:
- Exposing port binding errors so it can be properly handled.
- Capturing stderr to the logs when registering certificates.

### DIFF
--- a/python/tk_framework_desktopserver/errors.py
+++ b/python/tk_framework_desktopserver/errors.py
@@ -13,21 +13,21 @@ All custom exceptions that the browser integration can emit are here.
 """
 
 
-class BrowserIntegration(Exception):
+class BrowserIntegrationError(Exception):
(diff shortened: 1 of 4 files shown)


python/tk_framework_desktopserver/errors.py
```diff
@@ -9,12 +9,26 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
-All custom exceptions that the desktop integration can emit are here.
+All custom exceptions that the browser integration can emit are here.
 """
 
 
-class MissingCertificate(Exception):
+class BrowserIntegration(Exception):
```
robblau: We should follow the Python convention of having exception classes end in Error, so MissingCertificateError, PortBusyError, etc.

PR #6, opened 2015-10-02
For #31344, certificate generation from the command line

jfboismenu-adsk:
Introduces a new script that allows to create the certificates from the command line.
- `python certificates.py` will create it
- `python certificates.py --remove` will remove it
- adding `--debug` will send debug level logging to the console.

Also renamed the environment variables to configure the server. They are now SGTK_BROWSER_INTEGRATION_PORT and SGTK_BROWSER_INTEGRATION_CERTIFICATE.

Also, SGTK_BROWSER_INTEGRATION_WHITELIST was introduced to change the whitelist.

### DIFF
--- a/app/app.py
+++ b/app/app.py
@@ -62,10 +62,6 @@ if __name__ == '__main__':
     from tk_framework_desktopserver import Server, get_logger
     from tk_framework_desktopserver import shotgun_api
 
-    logger = get_logger()
-    logger.addHandler(logging.StreamHandler())
-    logger.setLevel(logging.DEBUG)
-
     parser = optparse.OptionParser()
     parser.add_option(
         "--debug", action="store_true", default=False,
@@ -94,6 +90,14 @@ if __name__ == '__main__':
 
     if options.block:
         print "Blocking calls for %s" % ", ".join(options.block)
+
+    logger = get_logger()
+    logger.addHandler(logging.StreamHandler())
+
+    if options.debug:
+        logger.setLevel(logging.DEBUG)
+    else:
+        logger.setLevel(logging.INFO)
 
     # Make a local copy of the handle because we are about to overwrite it.
     ShotgunAPI = shotgun_api.ShotgunAPI
--- a/app/app.py
+++ b/app/app.py
@@ -10,9 +10,9 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
+import os
 import sys
 import optparse
-import traceback
 
 def _parse_options():
     """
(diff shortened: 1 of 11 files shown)


app/app.py
```diff
@@ -109,8 +118,9 @@ def __getattr__(self, name):
 
     server = Server(
         debug=options.debug,
-        keys_path=os.environ.get("TANK_DESKTOP_INTEGRATION_CERTIFICATE", "../resources/keys"),
-        port=os.environ.get("TANK_DESKTOP_INTEGRATION_PORT", 9000)
+        keys_path=os.environ.get("SGTK_BROWSER_INTEGRATION_CERTIFICATE", "../resources/keys"),
+        port=os.environ.get("SGTK_BROWSER_INTEGRATION_PORT", 9000),
+        whitelist=os.environ.get("SGTK_BROWSER_INTEGRATION_WHITELIST")
```
robblau: Using individual env vars to override feels like overkill when we can use an env var to point at the config file.  Why not just have that and then read these values from there?  Feels easier to maintain.

app/certificates.py
```diff
@@ -0,0 +1,158 @@
...
+    # Check if there is a certificate registered with the keychain.
+    if cert_handler.is_registered():
+        logger.debug("Removing certificate from database.")
+        __warn_for_prompt()
+        cert_handler.unregister()
+        logger.info("The certificate is now unegistered.")
+    else:
+        logger.info("No certificate was registered.")
+
+    # Make sure the certificates exist.
```
robblau: Should this be "Make sure the certificates do not exist."?

app/certificates.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if options.debug is True:
+        logger.setLevel(logging.DEBUG)
+    else:
+        logger.setLevel(logging.INFO)
+
+    folder = os.environ.get("SGTK_BROWSER_INTEGRATION_CERTIFICATE", "../resources/keys")
+    folder = os.path.expanduser(folder)
+    folder = os.path.abspath(folder)
+    folder = os.path.normpath(folder)
```
robblau: Along with the env vars above, should this be a path to a config.ini that the server would also point to so that you don't need different mechanisms for pointing this to a cert and the server itself?

app/app.py
```diff
@@ -11,106 +11,51 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import sys
-import os
```
jfboismenu-adsk: I stripped a lot of code out of this file. A lot of it was not valid anymore and it didn't look like an officially supported tool, since there were comments about stuff that needs to be tested (which was out of date anyway).

python/tk_framework_desktopserver/settings.py
```diff
@@ -0,0 +1,135 @@
...
+
+import os
+import ConfigParser
+
+from . import logger
+
+logger = logger.get_logger("settings")
+
+
+class Settings(object):
```
jfboismenu-adsk: This class will be the basis of a refactor in tk-framework-desktopstartup to merge the settings from the desktop server and the desktop.

PR #7, opened 2015-10-13
For #31572, add missing message when removing certificate on MacOS

jfboismenu-adsk:
- User is now prompted when removing the certificate on MacOS.
- Fixed how the resources/keys path is built so it looks nicer when it is printed to the console.

### DIFF
--- a/app/settings.py
+++ b/app/settings.py
@@ -58,6 +58,6 @@ def get_settings(configuration):
         # into the resource files to look for the keys
         os.path.join(
             os.path.dirname(app_dir),
-            "resources/keys"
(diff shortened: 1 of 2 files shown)


app/settings.py
```diff
@@ -58,6 +58,6 @@ def get_settings(configuration):
         # into the resource files to look for the keys
         os.path.join(
             os.path.dirname(app_dir),
-            "resources/keys"
+            os.path.join("resources", "keys")
```
robblau: join takes an arbitrary number of paths... no need to add a nested call here.

PR #8, opened 2015-10-14
Fixes DUPLICATE_SAME_ACCESS issues on Windows

jfboismenu-adsk:
Bug and fix are documented here: https://bugs.python.org/issue3905

### DIFF
--- a/python/tk_framework_desktopserver/certificates.py
+++ b/python/tk_framework_desktopserver/certificates.py
@@ -136,14 +136,6 @@ class _CertificateHandler(object):
         return "Shotgun" in self._check_call(
             "validating if the certificate was installed", self._get_is_registered_cmd()
         )
-        )
-        stdout, _ = p.communicate()
-        if p.returncode != 0:
-            self._logger.error("Unexpected output:\n%s" % stdout)
-            raise CertificateRegistrationError(
-                "There was a problem validating if a certificate was installed."
-            )
-        return "Shotgun" in stdout
 
     def unregister(self):
         """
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/certificates.py
```diff
@@ -97,12 +97,32 @@ def _check_call(self, ctx, cmd):
...
+        # FIXME: This should be refactored to use the Command module from this framework.
+        # However this would mean a refactor that we can't afford at the moment so we'll
+        # stick with the minimal viable fix.
+
+        # Do not use popen.check_call because it won't redirect stderr to stdout properly
+        # and it can't close stdin which causes issues in certain configurations on Windows.
+        if sys.platform == "win32":
+            # More on this Windows specific fix here: https://bugs.python.org/issue3905
+            p = subprocess.Popen(
+                cmd, shell=True,
```
thebeeland: Is shell=True here necessary? If the primary use for this is certutil, it appears as though it's not necessary after a quick test. In this case it's not a huge deal because this is a private method and we can trust the input command, but in general it's best to not use it unless you're forced to.

jfboismenu-adsk: I used an array based approach originally but I felt it made the commands unreadable and harder to edit. Then I realized the inputs were safe therefore using shell=True wouldn't be too bad and then the command strings looked a lot cleaner.

thebeeland: In this case it's safe, so I'm fine with that if you prefer to go this route.

PR #9, opened 2015-10-28
For #33073, do not log sensitive information

jfboismenu-adsk:
When Twisted is used in debug mode, message content is logged. This can pose a security risk since that information will go to a log file on the user's drive. However this feature is very handy for debugging the server, so now Twisted's logging is sent to stdout and not propagated to the parent logger when debug is set. If debug is not set, the output of Twisted is considered safe and logger messages will be propagated to the parent accordingly.

This refactoring also has the benefit that now standard Twisted logging will be logged to desktop, so we can know if there are unexpected errors inside Twisted.

### DIFF
--- a/python/tk_framework_desktopserver/process_manager.py
+++ b/python/tk_framework_desktopserver/process_manager.py
@@ -152,8 +152,9 @@ class ProcessManager(object):
         has_error = return_code != 0
 
         if has_error:
-            raise Exception("{message_error}\nCommand: {command}\nReturn code: {return_code}\nOutput: {std_out}\nError: {std_err}"
-                            .format(message_error=message_error, command=args, return_code=return_code, std_out=out, std_err=err))
+            # Do not log the command line, it might contain sensitive information.
+            raise Exception("{message_error}\nReturn code: {return_code}\nOutput: {std_out}\nError: {std_err}"
+                            .format(message_error=message_error, return_code=return_code, std_out=out, std_err=err))
 
         return True
 
@@ -195,6 +196,7 @@ class ProcessManager(object):
 
             return (out, err, return_code)
         except Exception, e:
+            # call_cmd is not supposed to include sentitive information in the error message.
             raise Exception("Error executing toolkit command: " + e.message)
 
     def _add_action_output(self, actions, out, err, code):
--- a/python/tk_framework_desktopserver/process_manager.py
+++ b/python/tk_framework_desktopserver/process_manager.py
@@ -196,7 +196,8 @@ class ProcessManager(object):
 
             return (out, err, return_code)
         except Exception, e:
-            # call_cmd is not supposed to include sentitive information in the error message.
+            # call_cmd is not including sentitive information in the error message, so this won't
+            # either.
             raise Exception("Error executing toolkit command: " + e.message)
 
     def _add_action_output(self, actions, out, err, code):
--- a/python/tk_framework_desktopserver/server.py
+++ b/python/tk_framework_desktopserver/server.py
@@ -21,6 +21,8 @@ from autobahn.twisted.websocket import WebSocketServerFactory, listenWS
 
 from .errors import MissingCertificateError, PortBusyError
 
+from .logger import get_logger
+
 
 class Server:
     _DEFAULT_PORT = 9000
@@ -36,11 +38,26 @@ class Server:
         self._whitelist = whitelist or self._DEFAULT_WHITELIST
         self._debug = debug
 
-        self._logger_root = logging.getLogger("tk-framework-desktopserver")
+        self._logger_root = get_logger()
 
         if self._debug:
-            # This will take the Twisted logging and forward it to Python's logging.
-            self._observer = log.PythonLoggingObserver("tk-framework-desktopserver.twisted")
+            # The framework's log can be logged to disk by other apps and the twisted logging can contain
+            # sensitive information, so make sure logs don't propagate to the root handler.
+            # Simply output the twisted text to the console.
+            twisted = logging.getLogger("twisted")
+            handler = logging.StreamHandler()
+            formatter = logging.Formatter('%(asctime)s [%(name)s.%(levelname)s] %(message)s')
+            handler.setFormatter(formatter)
+            twisted.addHandler(handler)
+            twisted.warning("-" * 30)
+            twisted.warning(
+                "YOU ARE LOGGING TWISTED INTERNAL MESSAGES TO THE CONSOLE. MAKE SURE YOU CLOSE THE "
+                "CONSOLE AFTER RUNNING THE APPLICATION AS TWISTED CAN LOG SENSITIVE INFORMATION."
+            )
+            twisted.warning("-" * 30)
+
+            # This will take the Twisted logging and forward it to Python's logging
+            self._observer = log.PythonLoggingObserver("twisted")
             self._observer.start()
 
     def get_logger(self):
--- a/python/tk_framework_desktopserver/server.py
+++ b/python/tk_framework_desktopserver/server.py
@@ -29,7 +29,7 @@ class Server:
     _DEFAULT_KEYS_PATH = "../resources/keys"
     _DEFAULT_WHITELIST = "*.shotgunstudio.com"
 
-    def __init__(self, port=None, debug=True, whitelist=None, keys_path=None):
+    def __init__(self, port=None, debug=False, whitelist=None, keys_path=None):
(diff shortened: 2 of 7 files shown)


python/tk_framework_desktopserver/server.py
```diff
@@ -36,12 +38,37 @@ def __init__(self, port=None, debug=True, whitelist=None, keys_path=None):
         self._whitelist = whitelist or self._DEFAULT_WHITELIST
         self._debug = debug
 
-        self._logger_root = logging.getLogger("tk-framework-desktopserver")
+        self._logger_root = get_logger()
+
+        twisted = get_logger("twisted")
```
manneohrstrom: should this be named `twisted_logger` for clarity?

python/tk_framework_desktopserver/process_manager.py
```diff
@@ -195,6 +196,8 @@ def execute_toolkit_command(self, pipeline_config_path, command, args):
 
             return (out, err, return_code)
         except Exception, e:
+            # call_cmd is not including sentitive information in the error message, so this won't
```
manneohrstrom: This command feels unnecessary - the previous comments I felt were helpful, pointing out that we should be omitting something from the error message that otherwise would seem natural to include - thereby preventing future users to adding it (back) in and serve as an explanation as to why it isn't there. But this comment feels like we are sort of almost saying "there is nothing here to do so therefore we aren't doing anything" !

python/tk_framework_desktopserver/server.py
```diff
@@ -21,13 +21,15 @@
...
+from .logger import get_logger
+
 
 class Server:
     _DEFAULT_PORT = 9000
     _DEFAULT_KEYS_PATH = "../resources/keys"
     _DEFAULT_WHITELIST = "*.shotgunstudio.com"
 
-    def __init__(self, port=None, debug=True, whitelist=None, keys_path=None):
+    def __init__(self, port=None, debug=False, whitelist=None, keys_path=None):
```
manneohrstrom: Docstrings would be helpful - maybe outside of the scope of this PR?

python/tk_framework_desktopserver/server.py
```diff
@@ -36,12 +38,37 @@ def __init__(self, port=None, debug=True, whitelist=None, keys_path=None):
         self._whitelist = whitelist or self._DEFAULT_WHITELIST
         self._debug = debug
 
-        self._logger_root = logging.getLogger("tk-framework-desktopserver")
+        self._logger_root = get_logger()
```
manneohrstrom: I feel that the previous form of using `logging.getLogger()` is much clearer. This explicitly tells the reader of the code that this method comes from the logging module. By just saying `get_logger()`, it's not clear if this is an internal method or some external thing.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -60,10 +59,10 @@ def connectionLost(self, reason):
                 self._logger.info("Certificate error!")
                 StatusServerProtocol.serverStatus = StatusServerProtocol.SSL_CERTIFICATE_INVALID
             else:
-                self._logger.info("Connection lost!")
+                self._logger.info("Connection closed.")
```
manneohrstrom: Any reason this is not dispatching messages to the engine logger?

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -82,7 +81,7 @@ def onConnect(self, response):
             # response.origin: xyz.shotgunstudio.com
             domain_valid = self._is_domain_valid(response.origin)
         except:
-            self._logger.exception()
+            self._logger.exception("Unexpected error while losing connection.")
```
manneohrstrom: Should this message say something else for clarity? E.g. `Unexpected error while trying to determine the originating domain` or something?

python/tk_framework_desktopserver/server.py
manneohrstrom: Sorry, after looking at the rest of the code I understand! So ignore the previous comment :) Sorry!

Grepping the code quickly, it looks like `self._logger_root` is only used inside `Server.get_logger()` method. Could the member variable be removed completely?

python/tk_framework_desktopserver/server.py
```diff
@@ -21,13 +21,15 @@
 
 from .errors import MissingCertificateError, PortBusyError
 
+from .logger import get_logger
+
 
 class Server:
```
manneohrstrom: Now I am curious why we are not deriving from `object`? I think we should always do that and if we don't there should be a *big* comment explaining why.

python/tk_framework_desktopserver/server.py
```diff
@@ -36,12 +38,37 @@ def __init__(self, port=None, debug=True, whitelist=None, keys_path=None):
...
+        self._logger_root = get_logger()
+
+        twisted = get_logger("twisted")
 
         if self._debug:
-            # This will take the Twisted logging and forward it to Python's logging.
-            self._observer = log.PythonLoggingObserver("tk-framework-desktopserver.twisted")
-            self._observer.start()
+            # When running the server in debug mode, the twisted framework will print out data
+            # exchanged with the client. Unforutnately, there's no way to filter out that
```
manneohrstrom: Unforutnately (typo)

python/tk_framework_desktopserver/server.py
```diff
@@ -36,12 +38,37 @@ def __init__(self, port=None, debug=True, whitelist=None, keys_path=None):
...
+            # information from since it is being logged at the INFO level just like regular Twisted
+            # logs.
+
+            # Because the package information is sensitive, we will make sure logs don't propagate
+            # to the root handler.
+            twisted.propagate = False
+
+            # Print Twisted logs to the console.
+            handler = logging.StreamHandler()
```
manneohrstrom: This doesn't seem right to me. 

The whole idea with logging is to present a unified architecture of structured message passing so that a receiving client can determine what to do with that. The client (desktop, shell, some other future integration) should decide what to do with this info, log it to stdout, write it to a file, a syslog, a port etc. 

The library should merely dispatch logging information to the listener without making any assumptions about where this goes. Therefore, I don't think a library should ever define a handler like this. Instead, we should ensure that the information is passed in such a way that it is safe.

Our libraries should never echo anything to stdout/stderr. 

We want this potentially sensitive info to show when in debug mode but not anywhere else? 

Check this out: https://docs.python.org/2/howto/logging-cookbook.html#using-loggeradapters-to-impart-contextual-information

I think a better approach here would be to create a special logger instance for twisted that has the necessary business logic to understand when to discard messages and not (e.g. based on debug logging). Then leave all the formatting, output etc to the client side. All that we should do at this level is to basically cull out twisted messages depending on whether the debug flag is set or not.

manneohrstrom:
Some questions about the approach. Back to you JF!

python/tk_framework_desktopserver/process_manager.py
jfboismenu-adsk: I put the comment in because this is the only occasion in this file where we are taking text coming from an outside source and trusting to add it to our own error string. All other places where we are raising exception is it clear that we've not adding sensitive information to the exception. I felt it would be good to underline the fact here since we have to be careful of what we log and what we don't.

python/tk_framework_desktopserver/server.py
jfboismenu-adsk: Yeah, as I said the other day during my presentation, the framework needs a LOT of love.

jfboismenu-adsk: Nah, we should do it as we find them. As I said the other day, the framework needs a lot of love documentation wise so we'd better start now.

python/tk_framework_desktopserver/server_protocol.py
jfboismenu-adsk: There is no engine. There is not even a dependency on sgtk.

python/tk_framework_desktopserver/server_protocol.py
jfboismenu-adsk: Oops, copy-paste error, thanks!

python/tk_framework_desktopserver/server.py
jfboismenu-adsk: We should, it simply slipped through everyone's radar. Thanks for pointing it out.

manneohrstrom:
Looks sweeeet! :cake: :rice_cracker: :candy:

PR #11, opened 2015-11-10
For #31017, mounted volumes are missing on OSX.

jfboismenu:
On MacOS the QFileDialog hides all hidden files. Unfortunately this means that /Volumes is hidden. As a quick hack to unblock our clients, we'll add /Volumes to the sidebar.

### DIFF
--- a/python/tk_framework_desktopserver/sgtk_file_dialog.py
+++ b/python/tk_framework_desktopserver/sgtk_file_dialog.py
@@ -50,7 +50,7 @@ class SgtkFileDialog(QtGui.QFileDialog):
         if sys.platform == "darwin":
             sidebar_urls = self.sidebarUrls()
             if self._VOLUMES_URL not in sidebar_urls:
-                sidebar_urls.append("file:///Volumes")
+                sidebar_urls.append(self._VOLUMES_URL)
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/sgtk_file_dialog.py
```diff
@@ -42,6 +43,13 @@ def __init__(self, multi=False, *args, **kwargs):
         if treeview:
             treeview.setSelectionMode(selection_mode)
 
+        # FIXME: On MacOS the QFileDialog hides all hidden files. Unfortunately /Volumes is hidden.
+        # As a quick hack to unblock our clients, we'll add /Volumes to the sidebar.
+        if sys.platform == "darwin":
+            sidebarUrls = self.sidebarUrls()
+            sidebarUrls.append("file:///Volumes")
+            self.setSidebarUrls(sidebarUrls)
```
robblau: Awesome workaround!
Only comment is that we should check if that url (or equivalent?) is already there so that we don't duplicate it.

jfboismenu:
This is what our clients will see: 
screen shot 2015-11-10 at 2 01 11 pm

PR #13, opened 2016-06-27
For #36848, 0.18 compatible logging

jfboismenu:
Tweaks the logger name of the framework so that it gets picked up automatically when used inside a process using Core 0.18. This change is backwards compatible with the previous version of the Shotgun Desktop.

Added also a bit of logic to catch certificate issues on Windows and updated the logging of commands to the logs.

This will be integrated in the latest desktopstartup, but doesn't require a repackaging of the Desktop installer.

### DIFF
--- a/python/tk_framework_desktopserver/certificates.py
+++ b/python/tk_framework_desktopserver/certificates.py
@@ -106,6 +106,7 @@ class _CertificateHandler(object):
 
         # Do not use popen.check_call because it won't redirect stderr to stdout properly
         # and it can't close stdin which causes issues in certain configurations on Windows.
+        self._logger.debug("%s: %s" % (ctx.capitalize(), cmd))
         if sys.platform == "win32":
             # More on this Windows specific fix here: https://bugs.python.org/issue3905
             p = subprocess.Popen(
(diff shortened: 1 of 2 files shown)


python/tk_framework_desktopserver/certificates.py
```diff
@@ -119,8 +120,8 @@ def _check_call(self, ctx, cmd):
         else:
             p = subprocess.Popen(cmd, stderr=subprocess.STDOUT, stdout=subprocess.PIPE, shell=True)
         stdout, _ = p.communicate()
+        self._logger.debug("Stdout:\n%s" % stdout)
```
jfboismenu: Always log the output, we never know what kind of silent failures our clients might encounter.

python/tk_framework_desktopserver/certificates.py
```diff
@@ -106,6 +106,7 @@ def _check_call(self, ctx, cmd):
 
         # Do not use popen.check_call because it won't redirect stderr to stdout properly
         # and it can't close stdin which causes issues in certain configurations on Windows.
+        self._logger.debug("%s: %s" % (ctx.capitalize(), cmd))
```
robblau: Why capitalize?  I just got burned with a spurious capitalize call elsewhere... is the actual string so bad?  If not, lets leave it pure.

PR #15, opened 2017-02-01
For #40460, expose private API for Desktop

jfboismenu:
The settings class has been reimplemented in order to use the `sgtk.util.UserSettings` class when no location is passed in. `location` as a string is still supported for standalone server workflow.

### DIFF
--- a/python/tk_framework_desktopserver/settings.py
+++ b/python/tk_framework_desktopserver/settings.py
@@ -31,6 +31,10 @@ class Settings(object):
     """
 
     _BROWSER_INTEGRATION = "BrowserIntegration"
+    _PORT_SETTING = "port"
+    _LOW_LEVEL_DEBUG_SETTING = "low_level_debug"
+    _WHITELIST_SETTING = "whitelist"
+    _CERTIFICATE_FOLDER_SETTING = "certificate_folder"
 
     def __init__(self, location, default_certificate_folder):
         """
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/settings.py
```diff
@@ -39,13 +43,38 @@ def __init__(self, location, default_certificate_folder):
...
 
+        # Retrieve all the settings from the configuration file or the Tookit UserSettings object.
+        if location is not None:
+            config = self._load_config(location)
+            port = self._get_value(config, self._BROWSER_INTEGRATION, self._PORT_SETTING, int)
+            low_level_debug = bool(
+                self._get_value(config, self._BROWSER_INTEGRATION, self._LOW_LEVEL_DEBUG_SETTING, int)
+            )
+            whitelist = self._get_value(config, self._BROWSER_INTEGRATION, self._WHITELIST_SETTING)
+            certificate_folder = self._get_value(config, self._BROWSER_INTEGRATION, self._CERTIFICATE_FOLDER_SETTING)
```
manneohrstrom: maybe some of these longer lines could be broken up?

python/tk_framework_desktopserver/settings.py
```diff
@@ -39,13 +43,38 @@ def __init__(self, location, default_certificate_folder):
...
+            port = user_settings.get_integer_setting(self._BROWSER_INTEGRATION, self._PORT_SETTING)
+            low_level_debug = user_settings.get_boolean_setting(
+                self._BROWSER_INTEGRATION, self._LOW_LEVEL_DEBUG_SETTING
+            )
+            whitelist = user_settings.get_setting(self._BROWSER_INTEGRATION, self._WHITELIST_SETTING)
+            certificate_folder = user_settings.get_setting(self._BROWSER_INTEGRATION, self._CERTIFICATE_FOLDER_SETTING)
+
+        self._port = port or 9000
+        self._low_level_debug = low_level_debug or False
+        self._whitelist = whitelist or "*.shotgunstudio.com"
```
manneohrstrom: it would be great if we could define these defaults as constants too.

python/tk_framework_desktopserver/settings.py
```diff
@@ -39,13 +43,38 @@ def __init__(self, location, default_certificate_folder):
...
             is overridable for each app that can use this settings object.
         """
-        self._config = self._load_config(location)
+
         self._default_certificate_folder = default_certificate_folder
 
+        # Retrieve all the settings from the configuration file or the Tookit UserSettings object.
+        if location is not None:
+            config = self._load_config(location)
+            port = self._get_value(config, self._BROWSER_INTEGRATION, self._PORT_SETTING, int)
```
manneohrstrom: would it make sense to mimic the same interface we have for the user settings?

PR #16, opened 2017-04-03
For #41703, move websocket code into server

jfboismenu:
This code was all taken from the desktop startup framework. The logic is pretty much the same as the original one. The only change is that the browser integration `enabled` setting is now managed by the framework itself.

### DIFF
--- a/framework.py
+++ b/framework.py
@@ -9,6 +9,7 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import sgtk
+import sys
 import os
 import struct
 from sgtk.util import LocalFileStorageManager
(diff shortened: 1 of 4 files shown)


framework.py
```diff
@@ -9,32 +9,169 @@
...
+
+            self._server = self._tk_framework_desktopserver.Server(
+                port=self._settings.port,
+                low_level_debug=self._settings.low_level_debug,
+                whitelist=self._settings.whitelist,
+                keys_path=self._settings.certificate_folder
+            )
+
+            self._server.start()
+        except:
```
thebeeland: I'd prefer an `except Exception:` here. It'll catch anything we care about, and will allow things like SystemExit and KeyboardInterrupt to still bubble up. It's also strongly suggested by pep8.

framework.py
```diff
@@ -9,32 +9,169 @@
...
+        if self._server and self._server.is_running():
+            self._server.tear_down()
+
+    def __ensure_certificate_ready(self):
+        """
+        Ensures that the certificates are created and registered. If something is amiss, then the
+        certificates are regenerated.
 
-    def start_server(self, debug=False, start_reactor=True):
+        :returns: True is the certificate is ready, False otherwise.
```
thebeeland: I don't see where this is returning `False`

python/tk_framework_desktopserver/certificates.py
```diff
@@ -274,7 +274,7 @@ def register(self):
 
         :returns: True on success, False on failure.
```
thebeeland: This doesn't appear to be the case. Looks like no return and we get an exception on failure.

python/tk_framework_desktopserver/settings.py
```diff
@@ -140,6 +153,10 @@ def dump(self, logger):
         """
         Dumps all the settings into the logger.
         """
+        if self.integration_enabled is None:
```
thebeeland: Why/How would this be the case? Assuming it's possible, should it be? The property implies that you should expect boolean return, which feels better to me than boolean or maybe None.

framework.py
jfboismenu: Good point! Will fix.

jfboismenu:
These comments have been addressed as part of the merge back to `develop/zero_config`

PR #17, opened 2017-04-12
For 40784: Adds websockets protocol v2.

thebeeland:
This is more expansive than just 40784, and covers all desktopserver changes made thus far for websockets v2.

The ShotgunAPI has been broken out into v1 and v2 variants, and a factory function is provided to choose and instantiate the correct API for a given protocol version. The v1 API is almost entirely untouched, other than having been moved into a new location.

The v2 API provides `get_actions` and `execute_action` public RPC methods. All caching-related logic is managed by `get_actions`. All command execution is handled by `execute_action`.

The cache is populated by way of a subprocess that bootstraps SGTK into the tk-shotgun engine. A list of engine commands is pulled from the engine, munged a bit, and then cached into an sqlite database along with a unique identifier built from the pipeline configuration URI and entity type. A contents state hash is computed from a list of entities queried from the site (all Software entities, by default) and, if the pipeline config is mutable, the modtimes of all yml files found within.

Commands are run via a subprocess that bootstraps into the tk-shotgun engine and calling the associated engine command's callback.

### DIFF
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -93,6 +93,7 @@ class ServerProtocol(WebSocketServerProtocol):
             raise websocket.http.HttpException(403, "Domain origin was rejected by server.")
         else:
             self._logger.info("Connection accepted.")
+            self._wss_key = response.headers["sec-websocket-key"]
 
     def onMessage(self, payload, isBinary):
         """
@@ -161,6 +162,7 @@ class ServerProtocol(WebSocketServerProtocol):
                 protocol_version,
                 message_host,
                 self.process_manager,
+                wss_key=self._wss_key,
             )
         except Exception, e:
             message_host.report_error("Unable to get a ShotgunAPI object: %s" % e)
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -15,7 +15,7 @@ import datetime
 from urlparse import urlparse
 import OpenSSL
 
-import shotgun_api
+from .shotgun import get_shotgun_api
 from .message_host import MessageHost
 from .status_server_protocol import StatusServerProtocol
 from .process_manager import ProcessManager
@@ -31,8 +31,10 @@ class ServerProtocol(WebSocketServerProtocol):
     Server Protocol
     """
 
-    # Server protocol version
-    PROTOCOL_VERSION = 1
+    # Initial state is v2. This might change if we end up receiving a connection
+    # from a client at v1.
+    PROTOCOL_VERSION = 2
+    SUPPORTED_PROTOCOL_VERSIONS = (1, 2)
 
     def __init__(self):
         self._logger = get_logger()
@@ -108,9 +110,7 @@ class ServerProtocol(WebSocketServerProtocol):
         # Special message to get protocol version for this protocol. This message doesn't follow the standard
         # message format as it doesn't require a protocol version to be retrieved and is not json-encoded.
         if decoded_payload == "get_protocol_version":
-            data = {}
-            data["protocol_version"] = self.PROTOCOL_VERSION
-            self.json_reply(data)
+            self.json_reply(dict(protocol_version=self.PROTOCOL_VERSION))
             return
 
         # Extract json response (every message is expected to be in json format)
@@ -123,45 +123,65 @@ class ServerProtocol(WebSocketServerProtocol):
         message_host = MessageHost(self, message)
 
         # Check protocol version
-        if message["protocol_version"] != self.PROTOCOL_VERSION:
-            message_host.report_error("Error. Wrong protocol version [%s] " % self.PROTOCOL_VERSION)
+        if message["protocol_version"] not in self.SUPPORTED_PROTOCOL_VERSIONS:
+            message_host.report_error(
+                "Unsupported protocol version: %s " % message["protocol_version"]
+            )
             return
 
-        # Run each request from a thread, even tough it might be something very simple like opening a file. This
-        # will ensure the server is as responsive as possible. Twisted will take care of the thread.
-        reactor.callInThread(self._process_message, message_host, message)
+        # TODO: This isn't going to work if we have two different clients
+        # connecting at two different protocol versions. This is a temp
+        # hack to placate message_host.py. <jbee>
+        self.PROTOCOL_VERSION = message["protocol_version"]
 
-    def _process_message(self, message_host, message):
+        # Run each request from a thread, even though it might be something very simple like opening a file. This
+        # will ensure the server is as responsive as possible. Twisted will take care of the thread.
+        reactor.callInThread(
+            self._process_message,
+            message_host,
+            message,
+            message["protocol_version"],
+        )
+
+    def _process_message(self, message_host, message, protocol_version):
 
         # Retrieve command from message
         command = message["command"]
 
         # Retrieve command data from message
-        data = {}
-        if "data" in command:
-            data = command["data"]
-
+        data = command.get("data", dict())
         cmd_name = command["name"]
 
         # Create API for this message
         try:
             # Do not resolve to simply ShotgunAPI in the imports, this allows tests to mock errors
-            shotgun = shotgun_api.ShotgunAPI(message_host, self.process_manager)
+            api = get_shotgun_api(
+                protocol_version,
+                message_host,
+                self.process_manager,
+            )
         except Exception, e:
-            message_host.report_error("Error in loading ShotgunAPI. " + e.message)
+            message_host.report_error("Unable to get a ShotgunAPI object: %s" % e)
             return
 
         # Make sure the command is in the public API
-        if cmd_name in shotgun.public_api:
+        if cmd_name in api.PUBLIC_API_METHODS:
             # Call matching shotgun command
             try:
-                func = getattr(shotgun, cmd_name)
-                func(data)
+                func = getattr(api, cmd_name)
             except Exception, e:
-                message_host.report_error("Error! Could not execute function (possibly wrong command arguments) for [%s] -- %s"
-                                          % (cmd_name, e.message))
+                message_host.report_error(
+                    "Could not find API method %s: %s" % (cmd_name, e)
+                )
+            else:
+                try:
+                    func(data)
+                except Exception, e:
+                    message_host.report_error(
+                        "Method call failed for %s: %s" % (cmd_name, e)
+                    )
         else:
-            message_host.report_error("Error! Wrong Command Sent: [%s]" % cmd_name)
+            message_host.report_error("Command %s is not supported." % cmd_name)
 
     def report_error(self, message, data=None):
         """
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -161,7 +161,6 @@ class ServerProtocol(WebSocketServerProtocol):
                 protocol_version,
                 message_host,
                 self.process_manager,
-                self._semaphore,
             )
         except Exception, e:
             message_host.report_error("Unable to get a ShotgunAPI object: %s" % e)
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -14,6 +14,7 @@ import re
 import datetime
 from urlparse import urlparse
 import OpenSSL
+import threading
 
 from .shotgun import get_shotgun_api
 from .message_host import MessageHost
@@ -39,6 +40,7 @@ class ServerProtocol(WebSocketServerProtocol):
     def __init__(self):
         self._logger = get_logger()
         self.process_manager = ProcessManager.create()
+        self._semaphore = threading.Semaphore()
 
     def onClose(self, wasClean, code, reason):
         pass
@@ -159,6 +161,7 @@ class ServerProtocol(WebSocketServerProtocol):
                 protocol_version,
                 message_host,
                 self.process_manager,
+                self._semaphore,
             )
         except Exception, e:
             message_host.report_error("Unable to get a ShotgunAPI object: %s" % e)
@@ -169,17 +172,31 @@ class ServerProtocol(WebSocketServerProtocol):
             # Call matching shotgun command
             try:
                 func = getattr(api, cmd_name)
+                requires_sync = (cmd_name in api.SYNCHRONOUS_METHODS)
             except Exception, e:
                 message_host.report_error(
                     "Could not find API method %s: %s" % (cmd_name, e)
                 )
             else:
+                # If a method is expecting to be run synchronously we
+                # need to make sure that happens. An example of this is
+                # the get_actions method in v2 of the api, which might
+                # trigger a cache update. If that happens, we can't let
+                # multiple cache processes occur at once, because each
+                # is bootstrapping sgtk, which won't play well if multiple
+                # are occurring at the same time, all of which potentially
+                # copying/downloading files to disk in the same location.
+                if requires_sync:
+                    self._semaphore.acquire()
                 try:
                     func(data)
                 except Exception, e:
                     message_host.report_error(
                         "Method call failed for %s: %s" % (cmd_name, e)
                     )
+                finally:
+                    if requires_sync:
+                        self._semaphore.release()
         else:
             message_host.report_error("Command %s is not supported." % cmd_name)
 
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -190,8 +190,9 @@ class ServerProtocol(WebSocketServerProtocol):
                 try:
                     func(data)
                 except Exception, e:
+                    import traceback
                     message_host.report_error(
-                        "Method call failed for %s: %s" % (cmd_name, e)
+                        "Method call failed for %s: %s" % (cmd_name, traceback.format_exc())
                     )
                 finally:
                     if requires_sync:
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -34,13 +34,34 @@ class ServerProtocol(WebSocketServerProtocol):
 
     # Initial state is v2. This might change if we end up receiving a connection
     # from a client at v1.
-    PROTOCOL_VERSION = 2
     SUPPORTED_PROTOCOL_VERSIONS = (1, 2)
+    LOCK = threading.Lock()
 
     def __init__(self):
         self._logger = get_logger()
-        self.process_manager = ProcessManager.create()
-        self._semaphore = threading.Semaphore()
+        self._process_manager = ProcessManager.create()
+        self._protocol_version = 2
+
+    @property
+    def logger(self):
+        """
+        The log handler.
+        """
+        return self._logger
+
+    @property
+    def process_manager(self):
+        """
+        The protocol handler's associated process manager object.
+        """
+        return self._process_manager
+
+    @property
+    def protocol_version(self):
+        """
+        The protocol handler's currently-supported protocol version.
+        """
+        return self._protocol_version
 
     def onClose(self, wasClean, code, reason):
         pass
@@ -113,7 +134,7 @@ class ServerProtocol(WebSocketServerProtocol):
         # Special message to get protocol version for this protocol. This message doesn't follow the standard
         # message format as it doesn't require a protocol version to be retrieved and is not json-encoded.
         if decoded_payload == "get_protocol_version":
-            self.json_reply(dict(protocol_version=self.PROTOCOL_VERSION))
+            self.json_reply(dict(protocol_version=self._protocol_version))
             return
 
         # Extract json response (every message is expected to be in json format)
@@ -132,10 +153,7 @@ class ServerProtocol(WebSocketServerProtocol):
             )
             return
 
-        # TODO: This isn't going to work if we have two different clients
-        # connecting at two different protocol versions. This is a temp
-        # hack to placate message_host.py. <jbee>
-        self.PROTOCOL_VERSION = message["protocol_version"]
+        self._protocol_version = message["protocol_version"]
 
         # Run each request from a thread, even though it might be something very simple like opening a file. This
         # will ensure the server is as responsive as possible. Twisted will take care of the thread.
@@ -188,7 +206,7 @@ class ServerProtocol(WebSocketServerProtocol):
                 # are occurring at the same time, all of which potentially
                 # copying/downloading files to disk in the same location.
                 if requires_sync:
-                    self._semaphore.acquire()
+                    self.LOCK.acquire()
                 try:
                     func(data)
                 except Exception, e:
@@ -198,7 +216,7 @@ class ServerProtocol(WebSocketServerProtocol):
                     )
                 finally:
                     if requires_sync:
-                        self._semaphore.release()
+                        self.LOCK.release()
         else:
             message_host.report_error("Command %s is not supported." % cmd_name)
 
--- a/python/tk_framework_desktopserver/shotgun/__init__.py
+++ b/python/tk_framework_desktopserver/shotgun/__init__.py
@@ -1,0 +1,22 @@
+# Copyright (c) 2017 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+def get_shotgun_api(protocol_version, host, process_manager):
+    """
+
+    """
+    if protocol_version == 1:
+        from .api_v1 import ShotgunAPI
+    elif protocol_version == 2:
+        from .api_v2 import ShotgunAPI
+    else:
+        raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
+
+    return ShotgunAPI(host, process_manager)
--- a/python/tk_framework_desktopserver/shotgun/__init__.py
+++ b/python/tk_framework_desktopserver/shotgun/__init__.py
@@ -8,15 +8,41 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-def get_shotgun_api(protocol_version, host, process_manager):
-    """
+# We want class-level variables to persist, so we'll setup a cache here
+# that the factory functuon below populates.
+__API_CACHE = dict()
 
+def get_shotgun_api(protocol_version, host, process_manager, wss_key):
     """
-    if protocol_version == 1:
-        from .api_v1 import ShotgunAPI
-    elif protocol_version == 2:
-        from .api_v2 import ShotgunAPI
+    A factory function that returns an rpc API instance. The given
+    protocol version will be taken into account when determining which
+    API version to instantiate and return.
+
+    :param int protocol_version: The protocol version that the returned
+        API instance must support.
+    :param host: The message host.
+    :param process_manager: A process manager. This is only used by the
+        protocol verion 1 API, but provided to all APIs for the sake of
+        consistency.
+    :param str wss_key: The unique key associated with a WSS connection.
+        This key is provided by the autobahn libary's ConnectionRequest
+        object at the time a new WSS connection is made.
+
+    :returns: An RPC API instance appropriate for the given protocol
+        version.
+    """
+    global __API_CACHE
+
+    if protocol_version in __API_CACHE:
+        ShotgunAPI = __API_CACHE[protocol_version]
     else:
-        raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
+        if protocol_version == 1:
+            from .api_v1 import ShotgunAPI
+            __API_CACHE[protocol_version] = ShotgunAPI
+        elif protocol_version == 2:
+            from .api_v2 import ShotgunAPI
+            __API_CACHE[protocol_version] = ShotgunAPI
+        else:
+            raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
 
-    return ShotgunAPI(host, process_manager)
+    return ShotgunAPI(host, process_manager, wss_key)
--- a/python/tk_framework_desktopserver/shotgun/__init__.py
+++ b/python/tk_framework_desktopserver/shotgun/__init__.py
@@ -10,7 +10,8 @@
 
 # We want class-level variables to persist, so we'll setup a cache here
 # that the factory functuon below populates.
-__API_CACHE = dict()
+from . import api_v1
+from . import api_v2
 
 def get_shotgun_api(protocol_version, host, process_manager, wss_key):
     """
@@ -31,18 +32,9 @@ def get_shotgun_api(protocol_version, host, process_manager, wss_key):
     :returns: An RPC API instance appropriate for the given protocol
         version.
     """
-    global __API_CACHE
-
-    if protocol_version in __API_CACHE:
-        ShotgunAPI = __API_CACHE[protocol_version]
+    if protocol_version == 1:
+        return api_v1.ShotgunAPI(host, process_manager, wss_key)
+    elif protocol_version == 2:
+        return api_v2.ShotgunAPI(host, process_manager, wss_key)
     else:
-        if protocol_version == 1:
-            from .api_v1 import ShotgunAPI
-            __API_CACHE[protocol_version] = ShotgunAPI
-        elif protocol_version == 2:
-            from .api_v2 import ShotgunAPI
-            __API_CACHE[protocol_version] = ShotgunAPI
-        else:
-            raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
-
-    return ShotgunAPI(host, process_manager, wss_key)
+        raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
--- a/python/tk_framework_desktopserver/shotgun/__init__.py
+++ b/python/tk_framework_desktopserver/shotgun/__init__.py
@@ -8,7 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-def get_shotgun_api(protocol_version, host, process_manager):
+def get_shotgun_api(protocol_version, host, process_manager, semaphore):
     """
 
     """
@@ -19,4 +19,4 @@ def get_shotgun_api(protocol_version, host, process_manager):
     else:
         raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
 
-    return ShotgunAPI(host, process_manager)
+    return ShotgunAPI(host, process_manager, semaphore)
--- a/python/tk_framework_desktopserver/shotgun/__init__.py
+++ b/python/tk_framework_desktopserver/shotgun/__init__.py
@@ -8,7 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-def get_shotgun_api(protocol_version, host, process_manager, semaphore):
+def get_shotgun_api(protocol_version, host, process_manager):
     """
 
     """
@@ -19,4 +19,4 @@ def get_shotgun_api(protocol_version, host, process_manager, semaphore):
     else:
         raise RuntimeError("Unsupported protocol version: %s" % protocol_version)
 
-    return ShotgunAPI(host, process_manager, semaphore)
+    return ShotgunAPI(host, process_manager)
--- a/python/tk_framework_desktopserver/shotgun/api_v1.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v1.py
@@ -40,7 +40,7 @@ class ShotgunAPI(object):
     # Nothing for v1. This is used in v2.
     SYNCHRONOUS_METHODS = []
 
-    def __init__(self, host, process_manager, semaphore):
+    def __init__(self, host, process_manager, wss_key):
         """
         API Constructor.
         Keep initialization pretty fast as it is created on every message.
@@ -50,6 +50,7 @@ class ShotgunAPI(object):
         """
         self.host = host
         self.process_manager = process_manager
+        self.wss_key = wss_key
 
     def _handle_toolkit_output(self, out, err, return_code):
         """
--- a/python/tk_framework_desktopserver/shotgun/api_v1.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v1.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2013 Shotgun Software Inc.
+# Copyright (c) 2017 Shotgun Software Inc.
 #
 # CONFIDENTIAL AND PROPRIETARY
 #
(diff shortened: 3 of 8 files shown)


python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,185 @@
...
+                ),
+            )
+
+        engine.logger.debug("Engine commands processed.")
+        engine.logger.debug("Inserting commands into cache...")
+
+        # Connect to the database and get the hashes we need to include in
+        # the insert. Each of the lookups call out to the browser_integration
+        # core hook.
+        connection = sqlite3.connect(cache_file)
```
jfboismenu: `with sqlite3.connect(cache_file) as connection:` would yield code without `try: except`. I think you won't need to call commit either.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,185 @@
...
+    toolkit_mgr.plugin_id = "shotgun_toolkit_menu_caching"
+    toolkit_mgr.base_configuration = base_configuration
+
+    for pc in data["pipeline_configs"]:
+        pc["project"] = pc.get("project", project_entity)
+    import pprint
+    pprint.pprint(data["pipeline_configs"])
+    pcs = toolkit_mgr.sort_and_filter_configuration_entities(
+        project=project_entity,
+        entities=data["pipeline_configs"],
```
jfboismenu: Once again, I'm worried this means we're not going to get actions from the site level pipeline configurations.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def emit(self, record):
+        """
+        Forwards the record back to to js via the engine communicator.
+
+        :param record: The record to log.
+        """
+        super(_BootstrapLogHandler, self).emit(record)
+
+        # always flush to ensure its seen by the js process
```
jfboismenu: JS? Javascript? Might be worth spelling it out.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,185 @@
...
+        lookup_hash = hash_data[pc["id"]]["lookup_hash"]
+        contents_hash = hash_data[pc["id"]]["contents_hash"]
+
+        toolkit_mgr.pipeline_configuration = pc["id"]
+        engine = toolkit_mgr.bootstrap_engine(engine_name, entity=entity)
+        pc_descriptor = toolkit_mgr.resolve_descriptor(project_entity)
+
+        # Clean up the pre-bootstrap logger since we can use the engine's
+        # logger from here on out.
+        sgtk.LogManager().root_logger.removeHandler(log_handler)
```
jfboismenu: Is this because tk-shotgun always logs on the console? I think adding a comment to that effect would be nice.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,185 @@
...
+def cache(cache_file, data, base_configuration, hash_data, engine_name):
+    import sqlite3
+    import sgtk
+
+    entity = dict(type=data["entity_type"], id=data["entity_id"])
+    project_entity = dict(id=data["project_id"], type="Project")
+
+    # Setup the bootstrap manager.
+    toolkit_mgr = sgtk.bootstrap.ToolkitManager()
+    toolkit_mgr.plugin_id = "shotgun_toolkit_menu_caching"
```
jfboismenu: Why not use basic.shotgun? Somehow might want to manage their shotgun actions through a second config, just like you could manage your maya actions with a second config.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,185 @@
...
+        try:
+            commands_blob = sqlite3.Binary(
+                cPickle.dumps(commands, cPickle.HIGHEST_PROTOCOL)
+            )
+
+            # Since we're likely to be updating out-of-date cached data more
+            # often than we're going to be inserting new rows into the cache,
+            # we'll try an update first. If no rows were affected by the update,
+            # we move on to an insert.
+            cursor.execute(
```
jfboismenu: I think you could replace this with a single INSERT OR REPLACE call.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+            new_name,
+            new_path_linux,
+            new_path_mac,
+            new_path_windows,
+            sgtk.pipelineconfig_utils.is_localized(data["pc_root_path"]),
+        )
+
+    def _filter_by_project(self, actions, sw_entities, project):
+        """
+        Filters out any actions that aren't permitted for the given project
```
jfboismenu: I feel like this method requires a bit more documentation. If I understand correctly:
- We cache actions per context and descriptor.
- Since multiple projects could be using the same descriptor, we're hoping to be able to share a cache of actions between them.
- We pass to the launch app the special flag the allows it to enumerate all software instead of just for the current project.
- We cache that result
- Then this method here culls any actions that are not accessible for the project.

If that is the case, I think it warrants an explanation in the code, cause that is complicated (also, VERY COOL).

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+        :param list entities: A list of entity dictionaries to be included in the
+            hash computation.
+
+        :returns: hash value
+        :rtype: int
+        """
+        # We dump the entities out as json, sorting on keys to ensure 
+        # consistent ordering of data.
+        hashable_data = dict(
+            entities=json.dumps(entities, sort_keys=True, default=self.__json_default),
```
jfboismenu: As far as I could tell by looking at the implementation of json.dumps, the sorting of keys is recursive. You can probably do away with creating a single structure and calling sort_keys only once.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -123,45 +126,81 @@ def onMessage(self, payload, isBinary):
...
+            else:
+                # If a method is expecting to be run synchronously we
+                # need to make sure that happens. An example of this is
+                # the get_actions method in v2 of the api, which might
+                # trigger a cache update. If that happens, we can't let
+                # multiple cache processes occur at once, because each
+                # is bootstrapping sgtk, which won't play well if multiple
+                # are occurring at the same time, all of which potentially
+                # copying/downloading files to disk in the same location.
+                if requires_sync:
```
thebeeland: It's for a couple of reasons:

1. In many cases, we have a single row in the cache that can be used my most get_actions requests by the client. This is especially true for tk-config-basic setups. As such, it makes sense to allow first-in-first-out behavior, as many (sometimes all) subsequent get_actions calls will then be able to return cached data. If they're all running async at roughly the same time, we'll likely trigger multiple cache inserts/updates due to race conditions.
2. The sqlite database that acts as the data store for the cache can only handle a single writing connection at a time. We don't want a bunch of caching subprocesses to be kicked off async at the same time, causing lots of sqlite inserts/updates at the same time.

python/tk_framework_desktopserver/shotgun/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+# We want class-level variables to persist, so we'll setup a cache here
+# that the factory functuon below populates.
```
thebeeland: Because the factory function imports the api modules in its local scope, subsequent calls to the factory won't return instances of the same class definition unless those are promoted to the global scope somehow. I'm achieving that with a module-level cache.

python/tk_framework_desktopserver/shotgun/api_v1.py
```diff
@@ -26,11 +26,21 @@ class ShotgunAPI():
...
+        "executeToolkitCommand",
+        "executeTankCommand",
+        "pickFileOrDirectory",
+        "pickFilesOrDirectories",
+        "version",
+        "getProjectActions",
+    ]
+
+    # Nothing for v1. This is used in v2.
+    SYNCHRONOUS_METHODS = []
```
thebeeland: I've not only because I've ignored the v1 API, but we can certainly give it a shot! It's a simple test.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+import json
+import subprocess
+import tempfile
+import contextlib
+import json
+import fnmatch
+import datetime
+
+import sgtk
+from sgtk.commands.clone_configuration import clone_pipeline_configuration_html
```
thebeeland: I'm not familiar with SgtkCommand. I'll have a look, and if it's a quick change I'll go ahead and get it switched over. Otherwise I'll make a polish/tech ticket for it.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+    DATABASE_FORMAT_VERSION = 1
+
+    # Return codes.
+    SUCCESSFUL_LOOKUP = 0
+    UNSUPPORTED_ENTITY_TYPE = 2
+    CACHING_ERROR = 3
+    COMMAND_SUCCEEDED = 0
+    COMMAND_FAILED = 1
+
+    BASE_CONFIG_URI = os.environ.get(
```
thebeeland: Yeah, using it exactly that way. I have tk-config-basic cloned and hacked in various places, and to test zero config setups I want to use that config as the base during dev. I would say that if we want to we can remove before shipping this, though it doesn't bother me over much to leave it in there.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+        if env_name is None:
+            self.host.reply(dict(retcode=self.UNSUPPORTED_ENTITY_TYPE))
+            return
+
+        manager = sgtk.bootstrap.ToolkitManager()
+        manager.base_configuration = self.BASE_CONFIG_URI
+
+        pcs = manager.sort_and_filter_configuration_entities(
+            project=project_entity,
+            entities=data["pipeline_configs"],
```
thebeeland: The thought here was that because the action menu in SG already has the PC entities queried, that we could reuse that data instead of re-querying it. That being said, now that I've shifted over to querying things from Python (like Software entities, as an example), it might make more sense to do the same for PC entities.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+        config_data = dict()
+
+        for pc in pcs:
+            # The hash that acts as the key we'll use to look up our cached
+            # data will be based on the entity type and the pipeline config's
+            # descriptor uri. We can get the descriptor from the toolkit
+            # manager and pass that through along with the entity type from SG
+            # to the core hook that computes the hash.
+            manager.pipeline_configuration = pc["id"]
+            pc_descriptor = manager.resolve_descriptor(project_entity)
```
thebeeland: Good point, and a good bit of polish to take care of. I'll add a polish ticket for it now.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: I tried and it wasn't working properly, but I'll try again. I got frustrated with the lack of a genuine upsert command in sqlite, and stackoverflows I found pointed to a couple things that didn't actually work. I said screw it and went with the approach here just because I knew it would work.

python/tk_framework_desktopserver/shotgun/__init__.py
jfboismenu: Why can't you import them at the global scope? Are there any side-effects to importing them?

python/tk_framework_desktopserver/shotgun/api_v1.py
jfboismenu: The way that v1 works at the moment is that if you keep hitting refresh it will spin up more and more and more `tank`s that will all read the config. Doesn't need to be fixed right away, but might be nice to turn on if we think it could speed up the menu generation.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: You can basicaly ask the tk instance for the `clone_configuration` command. it's interface might be as nice as what you have for the HTML stuff however, hence why I said it might need to finessing.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: I have a single configuration in Shotgun that targets basic.* and uses a dev descriptor pointing to my repo on disk. It is being picked up by every project.
screen shot 2017-04-13 at 14 45 41

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: internal 42066

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: Only if the SG menu also returns pipeline configurations that are not assigned to a project would this work.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: And when I said Western, I was thinking Eastern :p (China, Japan, etc)

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
jfboismenu: It's not a big deal, the code is still quite readable.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: I think it might be premature optimization, to be honest. I'm going to switch it to use get_pipeline_configuration and cache the results based on wss connection key, the way I'm doing for software entities.

python/tk_framework_desktopserver/shotgun/__init__.py
thebeeland: Probably not. I'll give it a shot.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -123,45 +126,81 @@ def onMessage(self, payload, isBinary):
...
-            message_host.report_error("Error. Wrong protocol version [%s] " % self.PROTOCOL_VERSION)
+        if message["protocol_version"] not in self.SUPPORTED_PROTOCOL_VERSIONS:
+            message_host.report_error(
+                "Unsupported protocol version: %s " % message["protocol_version"]
+            )
             return
 
-        # Run each request from a thread, even tough it might be something very simple like opening a file. This
+        # TODO: This isn't going to work if we have two different clients
+        # connecting at two different protocol versions. This is a temp
```
thebeeland: Looking at this again, I believe it's just a matter of storing the protocol version on the object instead of the class. Each unique wss connection from a client will spin up its own protocol handler object, so in theory each connection could come from a different protocol version and work just fine.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: I've removed the entirety of the custom logging stuff. That had slowly been pushed out of the way as I developed this, and it turned out that there was only a single debug message being sent from the non-engine logger. No need to keep it as a result.

python/tk_framework_desktopserver/server_protocol.py
thebeeland: Yep, good to go now. This TODO is now gone as of the next push!

python/tk_framework_desktopserver/shotgun/__init__.py
thebeeland: Rearranged some logic and it works fine now with global imports of the api_v1 and api_v2 modules. The module-level variable has been removed as of the next push.

python/tk_framework_desktopserver/shotgun/api_v1.py
thebeeland: internal ticket 42071 for polish

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: internal ticket 42073 for polish

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: It seems to work great using manager.get_pipeline_configurations(). Next push here and in the shotgun repo will get rid of the pipeline configs being passed down with messages.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,758 @@
...
+                base_configuration=self.BASE_CONFIG_URI,
+                hash_data=hash_data,
+                engine_name=self.ENGINE_NAME,
+            )
+        )
+
+        output = None
+
+        try:
+            output = subprocess.check_output([sys.executable, script, args_file])
```
thebeeland: Are the potential problems involved related to the parsability of the output caught and returned by check_output? If so, since I'm not parsing the output in any way, aren't the hoops that the v1 api is jumping through unnecessary here? I'm just wanting the output in case I need to pass it to a logger.error call. What the actual output is that's caught is of little consequence.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -224,10 +224,9 @@ def get_actions(self, data):
 
         manager = sgtk.bootstrap.ToolkitManager()
         manager.base_configuration = self.BASE_CONFIG_URI
-
-        pcs = manager.sort_and_filter_configuration_entities(
```
jfboismenu: Does this mean we can remove `sort_and_filter_configuration_entities` from core? If this was the only call site for that method, I say kill it.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -653,9 +683,8 @@ def _init_db(self):
         # as UTF-8 (byte string) or unicode. And in the latter case, the returned data
         # will always be unicode.
         connection.text_factory = str
-        c = connection.cursor()
 
-        try:
+        with contextlib.closing(connection.cursor()) as c:
```
jfboismenu: I don't know why the cursor doesn't have `__enter__` `__exit__` implemented on them it seems such an obvious feature to add.
maybe I should submit a PR to cpython :p

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -19,7 +19,7 @@ def execute(config, project, name, entities, base_configuration, engine_name):
 
     # Setup the bootstrap manager.
     toolkit_mgr = sgtk.bootstrap.ToolkitManager()
-    toolkit_mgr.plugin_id = "shotgun_toolkit_command_execution"
+    toolkit_mgr.plugin_id = "basic.shotgun.execute"
```
jfboismenu: Shouldn't `basic.shotgun` be enough?

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: There's that, but there's also close_fds=True, which needs to be passed in. This is to avoid having child processes inherit the file descriptors from the parent process, which on Windows means that the port 9000 is considered in used as long as the desktop and any child process are running. The launch app on Windows actually blocks until the DCC is closed IIRC for certain DCCs IIRC, which caused this issue. At the very least I would recommend using Popen and doing something similar to [this](https://github.com/shotgunsoftware/tk-core/blob/develop/zero_config/python/tank/util/process.py#L57-L65) and pass in `close_fds=True`

Maybe we can start releasing with this bit only and see where that takes us.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sys
+import cPickle
+
+def cache(cache_file, data, base_configuration, hash_data, engine_name):
+    import sqlite3
```
manneohrstrom: it would be great with extensive debug logging here as the caching process starts up and executes!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+    import sgtk
+
+    entity = dict(type=data["entity_type"], id=data["entity_id"])
+    project_entity = dict(id=data["project_id"], type="Project")
+
+    # Setup the bootstrap manager.
+    toolkit_mgr = sgtk.bootstrap.ToolkitManager()
+    toolkit_mgr.plugin_id = "basic.shotgun.cache"
```
manneohrstrom: why not just `basic.shotgun`? maybe add a comment

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+    toolkit_mgr.base_configuration = base_configuration
+
+    pcs = toolkit_mgr.get_pipeline_configurations(
+        project=project_entity,
+    ) or [dict(id=None)]
```
manneohrstrom: this code is hard for me to read. maybe add a comment or rephrase it.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+
+    for pc in pcs:
```
manneohrstrom: future optimizations: if i am doing dev and have 20 dev sandboxes, and make a change to one of them, it will re-cache all of them. Maybe there could be a way to determine (via the core hooks) which one of them needs rebuilding?)

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+                ),
+            )
+
+        engine.logger.debug("Engine commands processed.")
+        engine.logger.debug("Inserting commands into cache...")
+
+        # Connect to the database and get the hashes we need to include in
+        # the insert. Each of the lookups call out to the browser_integration
+        # core hook.
+        with sqlite3.connect(cache_file) as connection:
```
manneohrstrom: what happens if the caching process fails? how is this communicated back to the main process and to the user?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+
+if __name__ == "__main__":
+    arg_data_file = sys.argv[1]
+
+    with open(arg_data_file, "r") as fh:
+        arg_data = cPickle.load(fh)
+
+    sys.path = arg_data["sys_path"]
+
+    cache(
```
manneohrstrom: love the simplicity of this file!

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -0,0 +1,80 @@
...
+import os
+
+def execute(config, project, name, entities, base_configuration, engine_name):
+    import sgtk
+
+    ms_flag = sgtk.platform.constants.LEGACY_MULTI_SELECT_ACTION_FLAG
+
+    # Setup the bootstrap manager.
+    toolkit_mgr = sgtk.bootstrap.ToolkitManager()
+    toolkit_mgr.plugin_id = "basic.shotgun.execute"
```
manneohrstrom: again, comment pls to explain

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+import fnmatch
+import datetime
+
+import sgtk
+from sgtk.commands.clone_configuration import clone_pipeline_configuration_html
+
+###########################################################################
+# Classes
+
+class ShotgunAPI(object):
```
manneohrstrom: naming nitpick: i find this name a little confusing. Perhaps `ShotgunWebsocketsAPI` or someting like that could be helpful to distinguish from the shotgun API? Feel free to ignore this comment :)

python/tk_framework_desktopserver/shotgun/api_v2.py
manneohrstrom: There is also `TK_BOOTSTRAP_CONFIG_OVERRIDE` which i think could work? it's the master dev override env var for bootstrapping.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+
+    ###########################################################################
+    # Properties
+
+    @property
+    def logger(self):
+        """
+        The associated engine's logger.
+        """
+        return self._engine.logger
```
manneohrstrom: again, consider logging cleanup.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+                self._clone_configuration(data)
+            except Exception as e:
+                self.host.reply(
+                    dict(
+                        retcode=self.COMMAND_FAILED,
+                        err=str(e),
+                        out=str(e),
+                    ),
+                )
+                raise
```
manneohrstrom: please add a comment here explaining why we are bubbling this up

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+
+        script = os.path.join(
+            os.path.dirname(__file__),
+            "scripts",
+            "execute_command.py"
+        )
+
+        output = None
+
+        try:
```
manneohrstrom: logging timings here would be great for support and debugging. can you use http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk.log.LogManager.log_timing ?

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -0,0 +1,80 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sys
+import cPickle
+import os
+
+def execute(config, project, name, entities, base_configuration, engine_name):
+    import sgtk
+
```
manneohrstrom: i suggest init logging to `tk-desktop` right away up here. 
http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk.log.LogManager.initialize_base_file_handler

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,125 @@
...
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sys
+import cPickle
+
+def cache(cache_file, data, base_configuration, hash_data, engine_name):
+    import sqlite3
+    import sgtk
```
manneohrstrom: i suggest init logging to `tk-desktop` right away up here. 
http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk.log.LogManager.initialize_base_file_handler

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+                    ),
+                )
+                return
+
+        project_entity, entities = self._get_entities_from_payload(data)
+
+        # It shouldn't ever happen at this stage of the workflow, but we'll
+        # check to make sure that we can get an environment for the type of
+        # we're dealing with here.
+        env_name = self.__pick_environment(project_entity, entities[0])
```
manneohrstrom: generally, more debugging could be helpful i think. (i just love debug!). I always tend to think through the case when someone is having issues and we ask them to send the `tk-shotgun.log` file and try to retrace their steps. In that case it's great to have a lot of output! Also, how do we turn on debug in the case of the websockets integration? Is there a simple way to do it?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+        # during a caching operation. This is going to be the situation when
+        # a page is pre-loading entity-level commands on page load, as opposed to
+        # passing down a specific entity that's been selected in an already-loaded
+        # page.
+        if "entity_id" in data and data["entity_id"] == -1:
+            if data["entity_type"] == "Project":
+                data["entity_id"] = data["project_id"]
+            else:
+                temp_entity = self._engine.shotgun.find_one(
+                    data["entity_type"],
```
manneohrstrom: will this be an issue with non-project entities? What if we wanted actions on the software entity? or the project entity?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+            # to the core hook that computes the hash.
+            manager.pipeline_configuration = pc["id"]
+            pc_descriptor = manager.resolve_descriptor(project_entity)
+            pc_key = pc_descriptor.get_uri()
+
+            pc_data = dict()
+            pc_data["contents_hash"] = self._get_contents_hash(
+                pc_descriptor,
+                self._get_site_state_data(),
+            )
```
manneohrstrom: debug logging for what is being returned from all hooks is helpful for troubleshooting i think.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+                                ),
+                                config=pc,
+                            )
+                        else:
+                            self.logger.debug(
+                                "Cache is out of date, recaching: %s %s" % (
+                                    contents_hash, cached_contents_hash
+                                )
+                            )
+                            self._cache_actions(data, config_data)
```
manneohrstrom: looks like we could pass in a list of pipeline config ids here maybe for performance? (future optimization) so we only re-cache what has changed.

python/tk_framework_desktopserver/shotgun/api_v2.py
manneohrstrom: there is also this in core: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/process.py#L27

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+        # the modtimes of all yml files found. When they're added to the digest,
+        # we'll dump as json sorting alphanumeric on yml file name to ensure
+        # consistent ordering of data.
+        if config_descriptor and config_descriptor.is_immutable() == False:
+            yml_files = dict()
+
+            # TODO: Deeper traversal that takes into account possible includes stuff
+            # based on environment variables, which this might not catch if the included
+            # file is outside of the config. <jbee>
+            for root, dir_names, file_names in os.walk(config_descriptor.get_path()):
```
manneohrstrom: for the future, maybe log a ticket to traverse using the API instead?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+                sort_keys=True,
+                default=self.__json_default,
+            )
+        )
+
+    def _get_entities_from_payload(self, data):
+        """
+        Extracts the relevant entity information from the given payload data.
+
+        :param dict data: The payload data.
```
manneohrstrom: an example data dump would be helpful

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+                pc_key,
+                project_entity,
+                data["entity_type"],
+            )
+            pc_data["descriptor"] = pc_descriptor
+            pc_data["entity"] = pc
+            config_data[pc["id"]] = pc_data
+
+        with self._db_connect() as (connection, cursor):
+            for pc in pcs:
```
manneohrstrom: nit: use `pipeline_config` and `pipeline_configs` for readability

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,780 @@
...
+        connection.
+
+        :returns: A list of Software entity dictionaries.
+        :rtype: list
+        """
+        if "software_entities" not in self.WSS_KEY_CACHE[self._wss_key]:
+            self.WSS_KEY_CACHE[self._wss_key]["software_entities"] = self._engine.shotgun.find(
+                "Software",
+                [],
+                fields=self._engine.shotgun.schema_field_read("Software").keys(),
```
manneohrstrom: performance: we know the fields, so can we hard code them? if not, please explain in a comment!

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Thanks for the explanation, JF. I'm going to ticket it and we'll get it sorted out sooner than later.

thebeeland: Tickets as internal 42123 and lined up for next sprint. Should be a quick one.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Most likely it does, yeah. Right now there's no use case for it since desktopserver no longer requires it.

internal ticket 42124 lined up for next sprint

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
thebeeland: Is there any reason to be concerned about a collision should an execute_command and a cache subprocess run and bootstrap at the same time? I don't know enough about how the plugins/ids work to know myself.

python/tk_framework_desktopserver/shotgun/api_v1.py
```diff
@@ -9,7 +9,7 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 
-class ShotgunAPI():
+class ShotgunAPI(object):
```
thebeeland: It would be tough to do that and maintain backwards compatibility with an older version of Shotgun. It would require adapting old-style RPC command names, which don't match what we consider convention now (mixed case instead of underscores), to new-convention method names.

I would rather see us deprecate support for the v1 API at some point in the future and just ditch the v1 API altogether.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
jfboismenu: Does the semaphone that you have prevent two bootstraps to happen at the same time? If that is the case, you won't have two bootstrapping happening at the same time. I think.

Ideas ? :)

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: I agree. I was just following the existing codebase. However, it's a quick change to make it something a bit more descriptive, so it's probably worth knocking it out along with other polish work.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: Internal ticket 42126 for a logging polish pass. That'll apply to most everything here you've mentioned related to logging. Lined up for next sprint.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: JF mentioned similar in his review. Polish ticket 42066 on internal

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: CalledProcessError is raised when the subprocess fails. This is caught and an error reply is sent to the client, where it's handled and displayed in the action menu.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
thebeeland: I'll be changing them to basic.shotgun. This was me misunderstanding the plugin_id, for the most part.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
jfboismenu: Also, we didn't have a valid use case before for returning the path and descriptor fields back with get_pipeline_configurations so we didn't include them back then. Now that we do, maybe it would make sense to return those as well? This would avoid a call back to the server to resolve everything.

@manneohrstrom , this would mean that the entities would now have fields `id`, `type`, `name`, `descriptor`, `path` (as a `ShotgunPath`). Does this sound reasonable to guys?

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
jfboismenu: Could we consider having the logs be sent directly to the main desktop process instead of a file? We would have to provide the RPC credentials. I'm worried too many writers to the file is going to cause an issue at some point. Especially on Windows. OR set up an endpoint on the websocket server for logging. (I like that a lot more actually!)

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
manneohrstrom: i think you are right @thebeeland - safer to have two processes like this - please add a comment to explain. nice one!

python/tk_framework_desktopserver/shotgun/api_v1.py
manneohrstrom: cool - just a thought! Maybe worth adding a comment to point that out?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
manneohrstrom: perfecto!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
manneohrstrom: i think so @jfboismenu. Would it be problematic for code that stashes the result into an sg dict? Maybe `(sg_dict, descriptor, path)= get_pipeline_configurations()`, where `sg_dict` is a std `{id, type, name}`?

jfboismenu: This method currently returns an array of dictionaries. It would still return the same thing, only now there would be two extra fields. I think the polish ticket should be about adding those two fields and that's it.

PR #19, opened 2017-04-21
For #42157: Supports legacy shotgun_*.yml environments.

thebeeland:
Also fixes incorrect usage of the TK_BOOTSTRAP_CONFIG_OVERRIDE environment variable.

In the event that the caching process finds shotgun_*.yml files in the config, it will make use of those rather than bootstrapping into the conventional environments. The intent is to maintain backwards compatibility.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -140,7 +140,7 @@ class ShotgunAPI(object):
                 project=project_entity,
                 sys_path=sys.path,
                 base_configuration=constants.BASE_CONFIG_URI,
-                engine_name=sgtk.platform.constants.SHOTGUN_ENGINE_NAME,
+                engine_name=constants.ENGINE_NAME,
             ),
         )
 
@@ -250,7 +250,7 @@ class ShotgunAPI(object):
             manager.pipeline_configuration = pc["id"]
             pc_descriptor = manager.resolve_descriptor(project_entity)
 
-            self.logger.debug("Resolved config descriptor: %s" % pc_descriptor)
+            self.logger.debug("Resolved config descriptor: %r" % pc_descriptor)
             pc_key = pc_descriptor.get_uri()
 
             pc_data = dict()
@@ -397,7 +397,7 @@ class ShotgunAPI(object):
                 sys_path=sys.path,
                 base_configuration=constants.BASE_CONFIG_URI,
                 hash_data=hash_data,
-                engine_name=sgtk.platform.constants.SHOTGUN_ENGINE_NAME,
+                engine_name=constants.ENGINE_NAME,
             )
         )
 
@@ -476,7 +476,14 @@ class ShotgunAPI(object):
         filtered = []
 
         for action in actions:
-            if "engine_name" not in action:
+            # The engine_name property of an engine command is defined by
+            # tk-multi-launchapp, and corresponds to the engine that provided
+            # the information necessary to register the launcher. If the action
+            # doesn't include that key, then it means the underlying engine
+            # command did not provide that property, and as such is not a
+            # launcher. Similarly, if it's set to None then the same applies
+            # and we don't need to test this action for filtering purposes.
+            if action.get("engine_name") is None:
                 continue
 
             # We're only interested in entities that are referring to the
@@ -489,7 +496,7 @@ class ShotgunAPI(object):
             for sw in associated_sw:
                 for sw_project in sw.get("projects", []):
                     if sw_project["id"] != project["id"]:
-                        self.logger.info("Action %s filtered out due to SW entity projects." % action)
+                        self.logger.debug("Action %s filtered out due to SW entity projects." % action)
                         filtered.append(action)
                         break
                 if action in filtered:
@@ -768,7 +775,7 @@ class ShotgunAPI(object):
                 self.logger.debug("Keeping command %s -- it has an associated app." % command)
                 filtered.append(command)
             else:
-                self.logger.info(
+                self.logger.debug(
                     "Command %s filtered out for browser integration." % command
                 )
 
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -52,10 +52,13 @@ class ShotgunAPI(object):
     COMMAND_SUCCEEDED = 0
     COMMAND_FAILED = 1
 
-    BASE_CONFIG_URI = os.environ.get(
-        "TK_BOOTSTRAP_CONFIG_OVERRIDE",
-        "sgtk:descriptor:app_store?name=tk-config-basic"
-    )
+    OVERRIDE_CONFIG_PATH = os.environ.get("TK_BOOTSTRAP_CONFIG_OVERRIDE")
+
+    if OVERRIDE_CONFIG_PATH is not None:
+        BASE_CONFIG_URI = "sgtk:descriptor:dev?path=%s" % OVERRIDE_CONFIG_PATH
+    else:
+        BASE_CONFIG_URI = "sgtk:descriptor:app_store?name=tk-config-basic"
+
     ENGINE_NAME = "tk-shotgun"
 
     def __init__(self, host, process_manager, wss_key):
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -23,6 +23,7 @@ import datetime
 import sgtk
 from sgtk.commands.clone_configuration import clone_pipeline_configuration_html
 from sgtk.util import process
+from . import constants
 
 ###########################################################################
 # Classes
@@ -42,24 +43,7 @@ class ShotgunAPI(object):
 
     # Stores data persistently per wss connection.
     WSS_KEY_CACHE = dict()
-
     DATABASE_FORMAT_VERSION = 1
-
-    # Return codes.
-    SUCCESSFUL_LOOKUP = 0
-    UNSUPPORTED_ENTITY_TYPE = 2
-    CACHING_ERROR = 3
-    COMMAND_SUCCEEDED = 0
-    COMMAND_FAILED = 1
-
-    OVERRIDE_CONFIG_PATH = os.environ.get("TK_BOOTSTRAP_CONFIG_OVERRIDE")
-
-    if OVERRIDE_CONFIG_PATH is not None:
-        BASE_CONFIG_URI = "sgtk:descriptor:dev?path=%s" % OVERRIDE_CONFIG_PATH
-    else:
-        BASE_CONFIG_URI = "sgtk:descriptor:app_store?name=tk-config-basic"
-
-    ENGINE_NAME = "tk-shotgun"
 
     def __init__(self, host, process_manager, wss_key):
         """
@@ -117,7 +101,7 @@ class ShotgunAPI(object):
             except Exception as e:
                 self.host.reply(
                     dict(
-                        retcode=self.COMMAND_FAILED,
+                        retcode=constants.COMMAND_FAILED,
                         err=str(e),
                         out=str(e),
                     ),
@@ -127,7 +111,7 @@ class ShotgunAPI(object):
                 self.logger.debug("Clone configuration successful.")
                 self.host.reply(
                     dict(
-                        retcode=self.COMMAND_SUCCEEDED,
+                        retcode=constants.COMMAND_SUCCEEDED,
                         err="",
                         out="",
                     ),
@@ -154,8 +138,8 @@ class ShotgunAPI(object):
                 entities=entities,
                 project=project_entity,
                 sys_path=sys.path,
-                base_configuration=self.BASE_CONFIG_URI,
-                engine_name=self.ENGINE_NAME,
+                base_configuration=constants.BASE_CONFIG_URI,
+                engine_name=sgtk.platform.constants.SHOTGUN_ENGINE_NAME,
             ),
         )
 
@@ -193,7 +177,7 @@ class ShotgunAPI(object):
         self.logger.debug("Command execution complete.")
         self.host.reply(
             dict(
-                retcode=self.COMMAND_SUCCEEDED,
+                retcode=constants.COMMAND_SUCCEEDED,
                 out="Command executed successfully.",
                 err="",
             ),
@@ -235,11 +219,11 @@ class ShotgunAPI(object):
         env_name = self.__pick_environment(project_entity, entity)
 
         if env_name is None:
-            self.host.reply(dict(retcode=self.UNSUPPORTED_ENTITY_TYPE))
+            self.host.reply(dict(retcode=constants.UNSUPPORTED_ENTITY_TYPE))
             return
 
         manager = sgtk.bootstrap.ToolkitManager(allow_config_overrides=False)
-        manager.base_configuration = self.BASE_CONFIG_URI
+        manager.base_configuration = constants.BASE_CONFIG_URI
         pcs = self._get_pipeline_configurations(
             manager,
             project_entity,
@@ -338,7 +322,7 @@ class ShotgunAPI(object):
                     self.host.reply(
                         dict(
                             err="Shotgun Desktop failed to get engine commands.",
-                            retcode=self.CACHING_ERROR,
+                            retcode=constants.CACHING_ERROR,
                             out="Caching failed!",
                         ),
                     )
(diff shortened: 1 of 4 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -52,10 +52,13 @@ class ShotgunAPI(object):
     COMMAND_SUCCEEDED = 0
     COMMAND_FAILED = 1
 
-    BASE_CONFIG_URI = os.environ.get(
-        "TK_BOOTSTRAP_CONFIG_OVERRIDE",
-        "sgtk:descriptor:app_store?name=tk-config-basic"
-    )
+    OVERRIDE_CONFIG_PATH = os.environ.get("TK_BOOTSTRAP_CONFIG_OVERRIDE")
```
manneohrstrom: Do we need to duplicate this logic here? why is it needed? is it to get a cache key? Could we instead add better interfaces to the toolkit manager so we can retrieve the bootstrap config uri?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -52,10 +52,13 @@ class ShotgunAPI(object):
...
+
+    if OVERRIDE_CONFIG_PATH is not None:
+        BASE_CONFIG_URI = "sgtk:descriptor:dev?path=%s" % OVERRIDE_CONFIG_PATH
+    else:
+        BASE_CONFIG_URI = "sgtk:descriptor:app_store?name=tk-config-basic"
```
manneohrstrom: would be great to break this out into a `constants.py` file

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -32,8 +34,14 @@ def cache(cache_file, data, base_configuration, hash_data, engine_name):
         contents_hash = hash_data[pc["id"]]["contents_hash"]
 
         toolkit_mgr.pipeline_configuration = pc["id"]
-        engine = toolkit_mgr.bootstrap_engine(engine_name, entity=entity)
         pc_descriptor = toolkit_mgr.resolve_descriptor(project_entity)
+        pc_path = pc_descriptor.get_path()
+
+        if glob.glob(os.path.join(pc_path, "config", "env", "shotgun_*.yml")):
```
manneohrstrom: related to - https://github.com/shotgunsoftware/tk-core/pull/444 - would be great if we could get this logic to be on the inside of core instead so we can change it more easily. 

I imagine something along the lines of:

- bootstrap via start engine()
- (note: important - the toolkit manager does not know anything about the shotgun engine - it's completely agnostic!)
- start engine runs and pick environment resolves into an environment which does not contain a shotgun engine.
- in that case, we kick into legacy mode and look for shotgun_xxxx stuff (old pathways)

Would something like that be feasible you think?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -70,6 +57,7 @@ def __init__(self, host, process_manager, wss_key):
         self._host = host
         self._engine = sgtk.platform.current_engine()
         self._wss_key = wss_key
+        self._logger = sgtk.platform.get_logger("api-v2")
```
manneohrstrom: we should avoid creating and passing loggers around. Instead, the pattern should be that whenever we need a logger, we create one (usually via `__name__` and [get_logger](http://developer.shotgunsoftware.com/tk-core/platform.html#sgtk.platform.get_logger). Instead of holding on to this logger like this, could we just put a `logger = sgtk.platform.get_logger(__name__)` at the top of the file? At least that's a step closer to where we want to be logging wise. 

It looks like the `self.logger` property is called from within the class itself a lot (weird pattern! - this old code is a little odd!) - we could switch all that round to just call `logger` instead. Is the `logger` property used by external classes as well? We should refactor that or put on the list of things to refactor in that logging ticket.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -151,8 +139,8 @@ def execute_action(self, data):
                 entities=entities,
                 project=project_entity,
                 sys_path=sys.path,
-                base_configuration=self.BASE_CONFIG_URI,
-                engine_name=self.ENGINE_NAME,
+                base_configuration=constants.BASE_CONFIG_URI,
+                engine_name=sgtk.platform.constants.SHOTGUN_ENGINE_NAME,
```
manneohrstrom: this is not part of the public core API, so in this case i would have kept the `ENGINE_NAME` constant in the class.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -232,30 +220,36 @@ def get_actions(self, data):
         env_name = self.__pick_environment(project_entity, entity)
 
         if env_name is None:
-            self.host.reply(dict(retcode=self.UNSUPPORTED_ENTITY_TYPE))
+            self.host.reply(dict(retcode=constants.UNSUPPORTED_ENTITY_TYPE))
             return
 
-        manager = sgtk.bootstrap.ToolkitManager()
-        manager.base_configuration = self.BASE_CONFIG_URI
+        manager = sgtk.bootstrap.ToolkitManager(allow_config_overrides=False)
```
manneohrstrom: (see comments in other PR about property)

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -232,30 +220,36 @@ def get_actions(self, data):
...
+        manager.base_configuration = constants.BASE_CONFIG_URI
         pcs = self._get_pipeline_configurations(
             manager,
             project_entity,
         ) or [dict(id=None, name="Primary")]
 
+        self.logger.debug("Pipeline configurations found: %s" % pcs)
```
manneohrstrom: can switch to a global `logger.debug(...)`

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -232,30 +220,36 @@ def get_actions(self, data):
...
+
         # We'll need to pass up the config names in order along with a dict of
         # pc_name => commands.
         pc_names = [pc["name"] for pc in pcs]
         all_actions = dict()
         config_data = dict()
 
         for pc in pcs:
+            self.logger.debug("Processing config: %s" % pc)
```
manneohrstrom: can switch to a global `logger.debug(...)`

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -232,30 +220,36 @@ def get_actions(self, data):
...
+
             # The hash that acts as the key we'll use to look up our cached
             # data will be based on the entity type and the pipeline config's
             # descriptor uri. We can get the descriptor from the toolkit
             # manager and pass that through along with the entity type from SG
             # to the core hook that computes the hash.
             manager.pipeline_configuration = pc["id"]
             pc_descriptor = manager.resolve_descriptor(project_entity)
+
+            self.logger.debug("Resolved config descriptor: %s" % pc_descriptor)
```
manneohrstrom: can switch to a global `logger.debug(...)`

manneohrstrom: i think if you use `%r` you get a string more suitable for debug

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -483,17 +475,20 @@ def _filter_by_project(self, actions, sw_entities, project):
         filtered = []
 
         for action in actions:
+            if "engine_name" not in action:
```
manneohrstrom: maybe add a comment here to explain why we are doign this

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -483,17 +475,20 @@ def _filter_by_project(self, actions, sw_entities, project):
...
+            associated_sw = [s for s in sw_entities if s["engine"] == action["engine_name"]]
 
             # Check the project against the projects list for matching Software
             # entities. If a Software entity's projects list is empty, then there
             # is no filtering to be done, as it's accepted by all projects.
             for sw in associated_sw:
                 for sw_project in sw.get("projects", []):
                     if sw_project["id"] != project["id"]:
-                        self.logger.debug("Action %s filtered due to SW entity projects." % action)
+                        self.logger.info("Action %s filtered out due to SW entity projects." % action)
```
manneohrstrom: will this be annoying for users who have their software organized per project?

python/tk_framework_desktopserver/shotgun/constants.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
+
+# RPC return codes.
```
manneohrstrom: are these for v2 of the API? In that case, maybe worth adding a comment about it

python/tk_framework_desktopserver/shotgun/constants.py
```diff
@@ -0,0 +1,26 @@
...
+SUCCESSFUL_LOOKUP = 0
+UNSUPPORTED_ENTITY_TYPE = 2
+CACHING_ERROR = 3
+COMMAND_SUCCEEDED = 0
+COMMAND_FAILED = 1
+
+# Pipeline configurations.
+OVERRIDE_CONFIG_PATH = os.environ.get("TK_BOOTSTRAP_CONFIG_OVERRIDE")
+
+if OVERRIDE_CONFIG_PATH is not None:
```
manneohrstrom: is this logic needed now that we have the new bootstrap parameter?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -19,8 +21,8 @@ def cache(cache_file, data, base_configuration, hash_data, engine_name):
     project_entity = dict(id=data["project_id"], type="Project")
 
     # Setup the bootstrap manager.
-    toolkit_mgr = sgtk.bootstrap.ToolkitManager()
-    toolkit_mgr.plugin_id = "basic.shotgun.cache"
+    toolkit_mgr = sgtk.bootstrap.ToolkitManager(allow_config_overrides=False)
```
manneohrstrom: ^ missing docstring for this method :)

manneohrstrom: this script should call [initialize_base_file_handler("tk-shotgun")](http://developer.shotgunsoftware.com/tk-core/utils.html#sgtk.log.LogManager.initialize_base_file_handler) as soon as it launches to make sure we capture as much of the log into to file as possible.

manneohrstrom: do we delete the pickle file with the command data?

manneohrstrom: similar comments for logging for the other script too!

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: All sounds good. I still have the logging polish ticket lined up for next sprint. I'll address this comment then.

The reason I made this change right now was just because I was looking for an easier way to parse the logs visually and via search.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Didn't mean to leave that in there. Switching it back to debug now.

python/tk_framework_desktopserver/shotgun/constants.py
thebeeland: Yes. The override environment variable will be used for the manager's base_configuration, but won't be allowed to override PipelineConfiguration entities. This logic here establishes what that base_configuration will be.

thebeeland:
Merging this. Manne has some comments to make, but they'll go in as polish separately.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -23,6 +23,7 @@
 import sgtk
 from sgtk.commands.clone_configuration import clone_pipeline_configuration_html
 from sgtk.util import process
+from . import constants
```
jfboismenu: I personally don't like this pattern of defining constants in a global file. This make code harder to update when cleaning stuff up. Until someone convinces me otherwise, that's one pattern I'd love for us to kill to be honest in Toolkit. I don't care enough about it right now to ask for changes, but to be honest if there's no benefit from moving things around I'd ague for leaving constants as they were.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -232,30 +220,37 @@ def get_actions(self, data):
...
+
             # The hash that acts as the key we'll use to look up our cached
             # data will be based on the entity type and the pipeline config's
             # descriptor uri. We can get the descriptor from the toolkit
             # manager and pass that through along with the entity type from SG
             # to the core hook that computes the hash.
             manager.pipeline_configuration = pc["id"]
             pc_descriptor = manager.resolve_descriptor(project_entity)
+
+            self.logger.debug("Resolved config descriptor: %r" % pc_descriptor)
```
jfboismenu: logging ❤️

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -483,17 +476,27 @@ def _filter_by_project(self, actions, sw_entities, project):
         filtered = []
 
         for action in actions:
+            # The engine_name property of an engine command is defined by
+            # tk-multi-launchapp, and corresponds to the engine that provided
+            # the information necessary to register the launcher. If the action
```
jfboismenu: Correct me if I am wrong, but you are testing for the presence of a setting from the launcher app to know if the actions are launchers? This seems brittle. If we're hardcoding this to something known, can't we at least hardcode it to app name insteand, `tk-multi-launchapp`? If we are reluctant of hardcoding an app name, maybe it needs to be a config setting or a framework hook?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -766,10 +772,11 @@ def _process_commands(self, commands, project, entities):
 
         for command in commands:
             if command["app_name"] is not None:
+                self.logger.debug("Keeping command %s -- it has an associated app." % command)
```
jfboismenu: Did you know you could use a comma instead of `%`when formatting the string?

`self.logger.debug("This %s my %s.", "is", "boomstick")`


This has two benefits:

The first, which isn't that important, is that formatting will be done only if the message needs to be written out. For debug messages, this means we can skip formatting and shave of some execution time. Not a big deal here, I know.

The second, which is more important, is that if you are missing some arguments in your `% (...)` the code will raise an exception. If you use comma separate the arguments, the Python logger will instead catch the exception itself and log an error saying which line in your code was missing some arguments. Considering we sometimes log stuff for things we don't have the greatest coverage for, its safer to log using commas knowing that if we never end up in that edge case and need to log something we can be certain a formatting error won't disrupt the app just because of a logging mistake.

thebeeland: I didn't know the logger did that for us. Thanks for the heads up.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: It's not just that it's identifying commands coming from launchapp, it's that we need the property in order perform the sw entity lookup at 491 below. Even if I used the app name to identify the command, I'd still have to check for the property. I don't really see it as flimsy...if the property isn't there then we don't know enough about command to be able to match it to an associated Software entity, and therefore have nothing left to do.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Note: I have a ticket in next sprint to do a cleanup/polish pass on logging for all of this. I'll take care of then when I do the rest of it for that ticket.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: I'll let you and @manneohrstrom duke that out. I moved things to constants.py after an earlier CR command from him. I'm not passionate enough about one approach over the other to make an argument for either.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: I completely misunderstood where `actions` was coming from. I thought you were iterating on the settings coming from the yml file. actions != env settings.
Sorry!!!

PR #21, opened 2017-04-26
#For 42140: Refactors components of wss v2.

thebeeland:
The intent of which is efficiency, cleanliness, and better handling of situations where a config entity in SG is unusable in an unrecoverable way. This should result in speed increases and stability against edge cases discovered during QA.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -121,17 +121,6 @@ class ShotgunAPI(object):
 
         project_entity, entities = self._get_entities_from_payload(data)
 
-        # It shouldn't ever happen at this stage of the workflow, but we'll
-        # check to make sure that we can get an environment for the type of
-        # we're dealing with here.
-        env_name = self.__pick_environment(project_entity, entities[0])
-
-        if env_name is None:
-            self.logger.error(
-                "The provided entity does not correspond to any environment."
-            )
-            return
-
         args_file = self._get_arguments_file(
             dict(
                 config=data["pc"],
@@ -217,16 +206,6 @@ class ShotgunAPI(object):
         # off the list.
         project_entity, entities = self._get_entities_from_payload(data)
         entity = entities[0]
-
-        # We can end up getting requests for actions for any entity type
-        # that exists in Shotgun. Many of those we don't want to provide
-        # commands for, so we can just notify the client that it's an
-        # unsupported entity type.
-        env_name = self.__pick_environment(project_entity, entity)
-
-        if env_name is None:
-            self.host.reply(dict(retcode=constants.UNSUPPORTED_ENTITY_TYPE))
-            return
 
         manager = sgtk.bootstrap.ToolkitManager()
         manager.allow_config_overrides = False
@@ -846,25 +825,6 @@ class ShotgunAPI(object):
             return item.isoformat()
         raise TypeError("Item cannot be serialized: %s" % item)
 
-    def __pick_environment(self, project_entity, entity):
-        """
-        Calls the pick_environment core hook on a context built from the
-        provided entity.
-
-        :param dict project_entity: The project entity dictionary.
-        :param dict entity: The entity dictionary.
-
-        :returns: The name of the associated environment, or None.
-        :rtype: str or None
-        """
-        return self._engine.sgtk.execute_core_hook(
-            sgtk.platform.constants.PICK_ENVIRONMENT_CORE_HOOK_NAME,
-            context=sgtk.Context(
-                self._engine.sgtk,
-                project=project_entity,
-                entity=entity,
-            )
-        )
 
 
 
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -225,57 +225,26 @@ class ShotgunAPI(object):
 
         manager = sgtk.bootstrap.ToolkitManager()
         manager.allow_config_overrides = False
+        manager.plugin_id = "basic.shotgun"
         manager.base_configuration = constants.BASE_CONFIG_URI
-        pcs = self._get_pipeline_configurations(
+
+        all_actions = dict()
+        all_pc_data = self._get_pipeline_configuration_data(
             manager,
             project_entity,
-        ) or [dict(id=None, name="Primary")]
-
-        self.logger.debug("Pipeline configurations found: %s" % pcs)
-
-        # We'll need to pass up the config names in order along with a dict of
-        # pc_name => commands.
-        pc_names = [pc["name"] for pc in pcs]
-        all_actions = dict()
-        config_data = dict()
-
-        for pc in pcs:
-            self.logger.debug("Processing config: %s" % pc)
-
-            # The hash that acts as the key we'll use to look up our cached
-            # data will be based on the entity type and the pipeline config's
-            # descriptor uri. We can get the descriptor from the toolkit
-            # manager and pass that through along with the entity type from SG
-            # to the core hook that computes the hash.
-            manager.pipeline_configuration = pc["id"]
-            pc_descriptor = manager.resolve_descriptor(project_entity)
-
-            self.logger.debug("Resolved config descriptor: %r" % pc_descriptor)
-            pc_key = pc_descriptor.get_uri()
-
-            pc_data = dict()
-            pc_data["contents_hash"] = self._get_contents_hash(
-                pc_descriptor,
-                self._get_site_state_data(),
-            )
-            pc_data["lookup_hash"] = self._get_lookup_hash(
-                pc_key,
-                project_entity,
-                data["entity_type"],
-            )
-            pc_data["descriptor"] = pc_descriptor
-            pc_data["entity"] = pc
-            config_data[pc["id"]] = pc_data
+            data,
+        )
 
         with self._db_connect() as (connection, cursor):
-            for pc in pcs:
+            for pc_id, pc_data in all_pc_data.iteritems():
+                pc = pc_data["entity"]
+
                 # The hash that acts as the key we'll use to look up our cached
                 # data will be based on the entity type and the pipeline config's
                 # descriptor uri. We can get the descriptor from the toolkit
                 # manager and pass that through along with the entity type from SG
                 # to the core hook that computes the hash.
-                pc_data = config_data[pc["id"]]
-                manager.pipeline_configuration = pc["id"]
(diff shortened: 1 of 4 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -621,6 +592,68 @@ def _get_lookup_hash(self, config_uri, project, entity_type):
...
+        :returns: A dictionary, keyed by PipelineConfiguration entity id, that
+            contains dictionaries with "contents_hash", "lookup_hash",
+            "descriptor", and "entity" keys.
+        :rtype: dict
+        """
+        if "config_data" in self.WSS_KEY_CACHE[self._wss_key]:
+            self.logger.debug("Pipeline config data found for %s", self._wss_key)
+        else:
+            config_data = dict()
+            pcs = self._get_pipeline_configurations(
```
manneohrstrom: do you think the code would be easier to read if this method was inline here instead?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -621,6 +592,68 @@ def _get_lookup_hash(self, config_uri, project, entity_type):
...
+                manager,
+                project_entity,
+            ) or [dict(id=None, name="Primary")]
+
+            for pc in pcs:
```
manneohrstrom: i suggest writing out: for `pipeline_config in pipeline_configs:`

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -621,6 +592,68 @@ def _get_lookup_hash(self, config_uri, project, entity_type):
...
+                # data will be based on the entity type and the pipeline config's
+                # descriptor uri. We can get the descriptor from the toolkit
+                # manager and pass that through along with the entity type from SG
+                # to the core hook that computes the hash.
+                manager.pipeline_configuration = pc["id"]
+
+                try:
+                    pc_descriptor = manager.resolve_descriptor(project_entity)
+                except sgtk.bootstrap.TankBootstrapError as exc:
+                    self.logger.debug("Unable to resolve config descriptor, skipping: %s" % pc)
```
manneohrstrom: how about a warning here instead?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -621,6 +592,68 @@ def _get_lookup_hash(self, config_uri, project, entity_type):
...
+
+                # The hash that acts as the key we'll use to look up our cached
+                # data will be based on the entity type and the pipeline config's
+                # descriptor uri. We can get the descriptor from the toolkit
+                # manager and pass that through along with the entity type from SG
+                # to the core hook that computes the hash.
+                manager.pipeline_configuration = pc["id"]
+
+                try:
+                    pc_descriptor = manager.resolve_descriptor(project_entity)
```
manneohrstrom: first we call `manager.get_pipeline_configurations()` and then next up we call `resolve_descriptor()` - for performance, could this be combined into a single call? This is looping over a bunch of entries so potentially slow if you have lots of configurations.

manneohrstrom: more importantly, why do we have to do all these checks at runtime like this?

Could we have a more basic check that does a single site check to get the state of pcs and software and if that comes out clean then BAM we just load everything up from the cache - and if not, we kick into 'expensive mode' where we begin figuring out what needs recaching and then kick of the process to handle all that. Could the single cache check be for example `the number of pc records + latest pc updated date + number of sw records + latest sw updated date`?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -621,6 +592,68 @@ def _get_lookup_hash(self, config_uri, project, entity_type):
...
+                pc_data["lookup_hash"] = self._get_lookup_hash(
+                    pc_key,
+                    project_entity,
+                    data["entity_type"],
+                )
+                pc_data["descriptor"] = pc_descriptor
+                pc_data["entity"] = pc
+                config_data[pc["id"]] = pc_data
+
+            self.WSS_KEY_CACHE[self._wss_key]["config_data"] = config_data
```
manneohrstrom: some more information about when this cache validates would be nice! both in the code, and i am also curious to understand more!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -225,57 +225,26 @@ def get_actions(self, data):
...
+                pc = pc_data["entity"]
+
                 # The hash that acts as the key we'll use to look up our cached
                 # data will be based on the entity type and the pipeline config's
                 # descriptor uri. We can get the descriptor from the toolkit
                 # manager and pass that through along with the entity type from SG
                 # to the core hook that computes the hash.
-                pc_data = config_data[pc["id"]]
-                manager.pipeline_configuration = pc["id"]
+                manager.pipeline_configuration = pc_id
```
jfboismenu: I think `manager.pipeline_configuration = pc_id` is dead code. I don't see it being used anywhere after this line.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: I believe there is an optimization ticket for adding descriptors into `get_pipeline_configurations`

thebeeland: internal #42066

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Maybe I don't understand, but how can that be possible? If that's the case, how does the manager know what config to bootstrap? It can't just operate off of the base_configuration all the time, unless I'm misunderstanding the fundamentals (or they've changed).

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Breaking it out into two lines and commenting it so it's easier to read and understand.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Duh. Ignore me. I misread the intent of JF's comment here. Definitely dead code...leftover after moving stuff around.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -13,10 +13,22 @@
...
+    :param list entities: The list of entities selected in the web UI when the
+        command action was triggered.
+    :param str base_configuration: The desired base pipeline configuration's
+        uri.
+    :param str engine_name: The name of the engine to bootstrap into. This
+        is most likely going to be "tk-shotgun"
+    """
+    # The local import of sgtk ensures that it occurs after sys.path is set
+    # to what the server sent over.
     import sgtk
```
jfboismenu: You are using sgtk to create the boostrapper and the using it reusing it after the bootstrap. I don't think this is actually causing issues in your code, but might I suggest moving everything from the `import sgtk` to `toolkit_mgr.bootstrap_engine` to another method so that the sgtk used to bootstrap is scoped to this method only?

Then you can `import sgtk`after the bootstrap happened and you are guaranteed you are using the new code.

jfboismenu: You are using sgtk to create the boostrapper and the using it reusing it after the bootstrap. I don't think this is actually causing issues in your code, but might I suggest moving everything from the `import sgtk` to `toolkit_mgr.bootstrap_engine` to another method so that the sgtk used to bootstrap is scoped to this method only?

Then you can `import sgtk`after the bootstrap happened and you are guaranteed you are using the new code.

Is `LEGACY_MULTI_SELECT_ACTION_FLAG` one of the values you said you couldn't access unless you added the new code to the desktop startup?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -12,9 +12,27 @@
...
+    :param str engine_name: The name of the engine to bootstrap into. This
+        is most likely going to be "tk-shotgun"
+    :param dict config_data: All relevant pipeline configuration data. This
+        dict is keyed by pipeline config entity id, each containing a dict
+        that contains, at a minimum, "entity", "lookup_hash", and
+        "contents_hash" keys.
+    """
+    # The local import of sgtk ensures that it occurs after sys.path is set
+    # to what the server sent over.
     import sgtk
```
jfboismenu: This bootstrapping is bug prone. You are bringing the core that bootstrap into the scope and the handle is still available after bootstrapping. For this method there's no real harm since you don't use sgtk after bootstrapping, but moving the instantiation of the manager to a separate method could avoid bugs in the future if we ever need to call methods on sgtk module after the bootstrap.

Also,  I just noticed that you are going multiple bootstraps in the loop. I am worried this might have some weird side-effects, but I don't have a solution to offer to be honest. We are using local imports in quite a few places in Toolkit, which means that the call from the manager might be doing some local imports which will then jump into the updated version of the library instead of the original one.

I think @manneohrstrom this qualifies for the worries we had a while back with the core swapping. :)

python/tk_framework_desktopserver/shotgun/api_v2.py
manneohrstrom: cool! let's talk about this one in the meeting! if nothing else just for my benefit! It's great that we have more optimized methods lined up - i think my basic question is if there is a tweak we can do which means we only need to do 2 shotgun find() queries (per shotgun page load) in the best case scenario on a fully warmed up cache!

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
manneohrstrom: maybe one we can touch on in the meeting too?

thebeeland: Thanks.

That variable isn't related, no. It was the allow_config_overrides property that I added to ToolkitManager.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: As discussed, this has been refactored such that a single config entity is handled by each call to cache_commands. I've also moved bootstrap routine into its own method, such that sgtk is scoped out afterwards.

PR #22, opened 2017-04-27
For adding some review comments

manneohrstrom:
Temp PR to make it easier to add comments and thoughts!

### DIFF
--- a/framework.py
+++ b/framework.py
@@ -26,11 +26,37 @@ class DesktopserverFramework(sgtk.platform.Framework):
         self._settings = None
         self._tk_framework_desktopserver = None
 
+    def add_different_user_requested_callback(self, cb):
+        """
+        Registers a callback to know when a different user or site is making browser integration requests.
+        The caller is not waiting for the callback to return.
+
+        :param function cb: Callback of the form:
+            def callback(site, user_id):
+                '''
+                Called when the site or user is different than the current site or user.
+
+                :param str site: Url of the site the request is coming from.
+                :param int user_id: Id of the HumanUser who made the request.
+                '''
+        """
+        # Lazy-init because engine is initialized after its frameworks, so QtCore is not initialized yet.
+        from sgtk.platform.qt import QtCore
+        if self._server:
+            self._server.notifier.different_user_requested.connect(cb, type=QtCore.Qt.QueuedConnection)
+
     ##########################################################################################
     # init and destroy
-    def launch_desktop_server(self):
+    def launch_desktop_server(self, host, user_id):
         """
         Initializes the desktop server.
+
+        The server actually supports two protocols, named v1 and v2. v1 can be used to process requests from any
+        users from any sites, while v2 can only be used to process requests from the currently authenticated
+        user.
+
+        :param str host: Host for which we desire to answer requests.
+        :param int user_id: Id of the user for which we desire to answer requests.
         """
         self._tk_framework_desktopserver = self.import_module("tk_framework_desktopserver")
 
@@ -67,10 +93,11 @@ class DesktopserverFramework(sgtk.platform.Framework):
             self.__ensure_certificate_ready()
 
             self._server = self._tk_framework_desktopserver.Server(
-                port=self._settings.port,
+                keys_path=self._settings.certificate_folder,
+                host=host,
+                user_id=user_id,
                 low_level_debug=self._settings.low_level_debug,
-                whitelist=self._settings.whitelist,
-                keys_path=self._settings.certificate_folder
+                port=self._settings.port
             )
 
             self._server.start()
--- a/framework.py
+++ b/framework.py
@@ -9,32 +9,166 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import sgtk
+import sys
 import os
+import struct
+from sgtk.util import LocalFileStorageManager
+
 
 class DesktopserverFramework(sgtk.platform.Framework):
+    """
+    Provides browser integration.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(DesktopserverFramework, self).__init__(*args, **kwargs)
+        self._server = None
+        self._settings = None
+        self._tk_framework_desktopserver = None
 
     ##########################################################################################
     # init and destroy
+    def launch_desktop_server(self):
+        """
+        Initializes the desktop server.
+        """
+        self._tk_framework_desktopserver = self.import_module("tk_framework_desktopserver")
 
-    def init_framework(self):
-        self.log_debug("%s: Initializing..." % self)
-        self.server = None
+        # Read the browser integration settings from disk. By passing in location=None, the Toolkit API will be
+        # used to locate the settings instead of looking at a specific file.
+        self._settings = self._tk_framework_desktopserver.Settings(
+            location=None,
+            default_certificate_folder=os.path.join(
+                LocalFileStorageManager.get_global_root(
+                    LocalFileStorageManager.CACHE, LocalFileStorageManager.CORE_V18
+                ),
+                "desktop",
+                "config",
+                "certificates"
+            )
+        )
+        self._settings.dump(self.logger)
+
+        # Twisted only runs on 64-bits.
+        if not self.__is_64bit_python():
+            self.logger.warning("The browser integration is only available with 64-bit versions of Python.")
+            self._integration_enabled = False
+        # Did the user disable it?
+        elif not self._settings.integration_enabled:
+            self.logger.info("Browser integration has been disabled in the Toolkit settings.")
+            self._integration_enabled = False
+        else:
+            self._integration_enabled = True
+
+        if not self._integration_enabled:
+            return
+
+        try:
+            self.__ensure_certificate_ready()
+
+            self._server = self._tk_framework_desktopserver.Server(
+                port=self._settings.port,
+                low_level_debug=self._settings.low_level_debug,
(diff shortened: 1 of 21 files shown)


framework.py
```diff
@@ -9,32 +9,166 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super(DesktopserverFramework, self).__init__(*args, **kwargs)
+        self._server = None
+        self._settings = None
+        self._tk_framework_desktopserver = None
 
     ##########################################################################################
     # init and destroy
+    def launch_desktop_server(self):
```
manneohrstrom: this forms an explicit contract with desktop. is this an old requirement (looks like something new that was put in)? Design wise, it would be great if desktop didn't know about the framework - could it just start up via `init`? Why do we need this method?

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+
+def execute(config, project, name, entities, base_configuration, engine_name):
+    import sgtk
+
+    ms_flag = sgtk.platform.constants.LEGACY_MULTI_SELECT_ACTION_FLAG
+
+    # Setup the bootstrap manager.
+    toolkit_mgr = sgtk.bootstrap.ToolkitManager()
+    toolkit_mgr.allow_config_overrides = False
```
manneohrstrom: it would be great if we could go through this use case again and why it's not working without this param. We should add comments to explain!

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -0,0 +1,81 @@
...
+    toolkit_mgr.plugin_id = "basic.shotgun"
+    toolkit_mgr.base_configuration = base_configuration
+
+    if config:
+        toolkit_mgr.pipeline_configuration = config.get("id")
+
+    if entities:
+        entity = entities[0]
```
manneohrstrom: for things like this, comments to elaborate will help future readers.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    command = engine.commands.get(name)
+
+    if not command:
+        msg = "Unable to find engine command: %s" % name
+        engine.logger.error(msg)
+        raise RuntimeError(msg)
+
+    props = command["properties"]
```
manneohrstrom: a lot of this business logic would be helped by some explanations!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,131 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sys
+import cPickle
+import glob
+import os
+
+def cache(cache_file, data, base_configuration, hash_data, engine_name):
+    import sqlite3
```
manneohrstrom: why the local import? comment would be helpful!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -0,0 +1,131 @@
...
+    import sgtk
+
+    entity = dict(type=data["entity_type"], id=data["entity_id"])
+    project_entity = dict(id=data["project_id"], type="Project")
+
+    # Setup the bootstrap manager.
+    toolkit_mgr = sgtk.bootstrap.ToolkitManager()
+    toolkit_mgr.allow_config_overrides = False
```
manneohrstrom: same here, talking through this use case would helpful for me!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,836 @@
...
+                    ),
+                )
+                return
+
+        project_entity, entities = self._get_entities_from_payload(data)
+
+        # It shouldn't ever happen at this stage of the workflow, but we'll
+        # check to make sure that we can get an environment for the type of
+        # we're dealing with here.
+        env_name = self.__pick_environment(project_entity, entities[0])
```
manneohrstrom: this one is a little odd to me - i don't think we should worry about this at all in the server fw. Do we need to? I don't think the desktop server should know about pick environment.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,836 @@
...
+                data["entity_id"] = temp_entity["id"]
+
+        project_entity, entities = self._get_entities_from_payload(data)
+        entity = entities[0]
+
+        # We can end up getting requests for actions for any entity type
+        # that exists in Shotgun. Many of those we don't want to provide
+        # commands for, so we can just notify the client that it's an
+        # unsupported entity type.
+        env_name = self.__pick_environment(project_entity, entity)
```
manneohrstrom: same here!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,836 @@
...
+        self._wss_key = wss_key
+        self._logger = sgtk.platform.get_logger("api-v2")
+
+        if self._wss_key not in self.WSS_KEY_CACHE:
+            self.WSS_KEY_CACHE[self._wss_key] = dict()
+
+        # Cache path on disk.
+        self._cache_path = os.path.join(
+            self._engine.cache_location, 
+            "shotgun_engine_commands_v%s.sqlite" % self.DATABASE_FORMAT_VERSION,
```
manneohrstrom: good practice!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,836 @@
...
+        # If we're on Windows, we'll want to stop the cmd.exe window
+        # from flashing on/off when our subprocesses are run.
+        if sys.platform == "win32":
+            si = subprocess.STARTUPINFO()
+            si.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+            kwargs["startupinfo"] = si
+
+        return kwargs
+
+    def _init_db(self):
```
manneohrstrom: should this be in the caching thread, since this is the place where the db is written to? that keeps a clean writer/reader relationship between the caching code (this code) and the script that does the caching...

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -0,0 +1,836 @@
...
+                            actions = self._process_commands(
+                                commands=cPickle.loads(str(cached_data[0])),
+                                project=project_entity,
+                                entities=entities,
+                            )
+                            self.logger.debug("Actions found in cache: %s" % actions)
+                            all_actions[pc["name"]] = dict(
+                                actions=self._filter_by_project(
+                                    actions,
+                                    self._get_software_entities(),
```
manneohrstrom: this feels like it should be done in the cache process? if we just use the sw entity as part of the invalidation key, then we could capture this at cache time instead perhaps? i guess i am wondering if we can move as much of the 'compute' as possible into the caching part and all that the main thread is doing is basically figuring out 'am i still up to date' and if the anwer is no, then fire up the cache script that does all the computation

thebeeland: Not sure I understand how/why this would go into the caching script. We're specifically not caching based on a specific project, and instead keying on config descriptor and entity type. As such, the project-level filtering has to occur at runtime and not at cache time.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: No need for sqlite3. I'll move it up to global.

framework.py
thebeeland: This is a question for @jfboismenu

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: What's the expected flow when the client requests actions for a HumanUser entity? Do we through through the entire caching process, detect it there, and store an empty action list for that entity type? It seems like a log of overhead to me.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: This should be doable, and safely. The get_actions RPC methods are executed synchronously, so there shouldn't be any risk of race conditions where multiple subprocesses all try to initialize the base state of the database.

I'll move it over!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: We should discuss this in person. I feel like I've typed it out enough at this point. ;)

framework.py
jfboismenu: There are two reasons for doing this.
Frameworks are always initialized before the engine that uses them. This cause the styling of the widget to be the boring Qt default instead of the sleek Desktop one. So this is getting called after the desktop has set its look and feel.

Second, you have to remember that this is the first step of the desktop server integration into the engine. In the second step, the desktop will be aware of the server because it will need to install callbacks to restart the app or reauthenticate when requests are coming from another site.

PR #24, opened 2017-05-04
For #42066: Gets config descriptors in bulk from ToolkitManager.

thebeeland:
Config descriptors are no longer queried one at a time. They come in bulk from ToolkitManager by way of `get_pipeline_configurations()`.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -677,7 +677,7 @@ class ShotgunAPI(object):
 
                 if pc_descriptor is None:
                     logger.warning(
-                        "Unable to resolve config descriptor, skipping: %s",
+                        "Unable to resolve config descriptor, skipping: %r",
                         pipeline_config,
                     )
                     continue
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -19,6 +19,7 @@ import contextlib
 import json
 import fnmatch
 import datetime
+import copy
 
 import sgtk
 from sgtk.commands.clone_configuration import clone_pipeline_configuration_html
@@ -237,7 +238,16 @@ class ShotgunAPI(object):
                 # descriptor uri. We can get the descriptor from the toolkit
                 # manager and pass that through along with the entity type from SG
                 # to the core hook that computes the hash.
-                pc_descriptor = pc_data["descriptor"]
+                pc_descriptor = pipeline_config["descriptor"]
+
+                # Since the config entity is going to be passed up as part of the
+                # reply to the client, we need to filter out the descriptor object.
+                # It's neither useful to the client, nor json encodable. We're
+                # copying the dict first, since the original is stored in an
+                # in-memory cache and is likely to be reused.
+                pipeline_config = copy.copy(pipeline_config)
+                del pipeline_config["descriptor"]
+
                 lookup_hash = pc_data["lookup_hash"]
                 contents_hash = pc_data["contents_hash"]
                 cached_data = []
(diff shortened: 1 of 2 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -237,7 +238,16 @@ def get_actions(self, data):
...
                 # to the core hook that computes the hash.
-                pc_descriptor = pc_data["descriptor"]
+                pc_descriptor = pipeline_config["descriptor"]
+
+                # Since the config entity is going to be passed up as part of the
+                # reply to the client, we need to filter out the descriptor object.
+                # It's neither useful to the client, nor json encodable. We're
+                # copying the dict first, since the original is stored in an
+                # in-memory cache and is likely to be reused.
+                pipeline_config = copy.copy(pipeline_config)
```
jfboismenu: Wouldn't it be simpler if the in-memory cache was the one who cloned it's results instead? This way you wouldn't have to worry about people corrupting the cache. I doubt that's going to be a performance issue either.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -381,9 +391,14 @@ def _cache_actions(self, data, config_data):
         arg_config_data = dict(
             lookup_hash = config_data["lookup_hash"],
             contents_hash=config_data["contents_hash"],
-            entity=config_data["entity"],
+            entity=copy.copy(config_data["entity"]),
         )
 
+        # If we have a descriptor, we don't want to pickle that. It's not
+        # needed in either the caching or execution scripts.
+        if "descriptor" in arg_config_data["entity"]:
```
jfboismenu: Why would a configuration not have a descriptor?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -391,13 +388,12 @@ def _cache_actions(self, data, config_data):
         arg_config_data = dict(
             lookup_hash = config_data["lookup_hash"],
             contents_hash=config_data["contents_hash"],
-            entity=copy.copy(config_data["entity"]),
+            entity=config_data["entity"],
         )
 
         # If we have a descriptor, we don't want to pickle that. It's not
         # needed in either the caching or execution scripts.
```
jfboismenu: If the descriptor is always present, then you can probably do away with the "If we have a descriptor" bit.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -716,7 +712,7 @@ def _get_pipeline_configuration_data(self, manager, project_entity, data):
             else:
                 cache["config_data"] = config_data
 
-        return cache["config_data"][entity_type]
+        return copy.deepcopy(cache["config_data"][entity_type])
```
jfboismenu: Could you add a comment explaining why you are cloning the cache. Thanks!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -391,13 +388,12 @@ def _cache_actions(self, data, config_data):
...
             contents_hash=config_data["contents_hash"],
-            entity=copy.copy(config_data["entity"]),
+            entity=config_data["entity"],
         )
 
         # If we have a descriptor, we don't want to pickle that. It's not
         # needed in either the caching or execution scripts.
-        if "descriptor" in arg_config_data["entity"]:
-            del arg_config_data["entity"]["descriptor"]
+        del arg_config_data["entity"]["descriptor"]
```
jfboismenu: If the descriptor is always present, then you can probably do away with the "If we have a descriptor" bit.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -780,7 +776,7 @@ def _get_site_state_data(self):
         else:
             logger.debug("Cached site state data found for %s", self._wss_key)
 
-        return self.WSS_KEY_CACHE[self._wss_key]["site_state_data"]
+        return copy.deepcopy(self.WSS_KEY_CACHE[self._wss_key]["site_state_data"])
```
jfboismenu: Same here.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -805,7 +801,7 @@ def _get_software_entities(self):
         else:
             logger.debug("Cached software entities found for %s", self._wss_key)
 
-        return self.WSS_KEY_CACHE[self._wss_key]["software_entities"]
+        return copy.deepcopy(self.WSS_KEY_CACHE[self._wss_key]["software_entities"])
```
jfboismenu: And here.

PR #26, opened 2017-05-06
For #42437: Task actions are now cached per entity instead of by type.

thebeeland:
* This allows for different actions to be configured for Tasks linked to Shots vs. those linked to Assets.

* Also adds local file linking file browse/pick RPC methods to api_v2.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -50,6 +50,13 @@ class ShotgunAPI(object):
     WSS_KEY_CACHE = dict()
     DATABASE_FORMAT_VERSION = 1
 
+    # Keys for the in-memory cache.
+    TASK_PARENT_TYPES = "task_parent_types"
+    PIPELINE_CONFIGS = "pipeline_configurations"
+    SITE_STATE_DATA = "site_state_data"
+    CONFIG_DATA = "config_data"
+    SOFTWARE_ENTITIES = "software_entities"
+
     def __init__(self, host, process_manager, wss_key):
         """
         API Constructor.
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -829,6 +889,29 @@ def _get_subprocess_kwargs(self):
...
+            logger.debug("Parent entity type found in cache for Task %s.", task_id)
+        else:
+            context = sgtk.context.from_entity(
+                self._engine.sgtk,
+                "Task",
+                task_id,
+            )
+            entity_type = context.entity["type"]
+
+            if "task_parent_types" in cache:
```
jfboismenu: Depending on whether you think the following is cleaner or not, you could also have written this as

```python
cache.setdefault("task_parent_types", {})[task_id] = entity_type
```

jfboismenu: ```setdefault``` retrieves the value for `task_parent_types`. If it doesn't exist, then it sets `{}` as the value and returns it.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -829,6 +889,29 @@ def _get_subprocess_kwargs(self):
 
         return kwargs
 
+    def _get_task_parent_entity_type(self, task_id):
+        """
+        Gets the Task entity's parent entity type.
+        """
+        cache = self.WSS_KEY_CACHE[self._wss_key]
+
+        if "task_parent_types" in cache and task_id in cache["task_parent_types"]:
```
jfboismenu: TASK_PARENT_TYPES = "task_parent_types" at the top of the file would be appreciated. This is non standard enough that it is worth constantifying.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -245,10 +255,23 @@ def get_actions(self, data):
                 # It's neither useful to the client, nor json encodable.
                 del pipeline_config["descriptor"]
 
-                lookup_hash = pc_data["lookup_hash"]
+                # In all cases except for Task entities, we'll already have a
+                # lookup hash computed. If we're dealing with a Task, though,
+                # it'll be a None and we'll need to compute is live. This is
```
jfboismenu: we'll need to compute _IT_ live?

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Switching all of the cache keys to constants. Good idea.

thebeeland:
Good nitpicks. Taken care of!

PR #27, opened 2017-05-09
For #42438: Adds a whitelist of entity types that are supported by wss2.

thebeeland:
The list is built from a set of constant entity types, plus entity types that the PublishedFile entity is configured to link to in the current site.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -687,6 +687,13 @@ class ShotgunAPI(object):
         :returns: A set of string entity types.
         :rtype: set
         """
+        if project_id is None:
+            logger.debug("Project id is None, looking up site schema.")
+            project_entity = None
+        else:
+            project_entity = dict(type="Project", id=project_id)
+            logger.debug("Looking up schema for project %s", project_entity)
+
         cache_is_initialized = self.ENTITY_TYPE_WHITELIST in self._cache
         project_in_cache = project_id in self._cache.get(self.ENTITY_TYPE_WHITELIST, dict())
 
@@ -695,7 +702,7 @@ class ShotgunAPI(object):
             schema = self._engine.shotgun.schema_field_read(
                 constants.PUBLISHED_FILE_ENTITY,
                 field_name="entity",
-                project_entity=dict(type="Project", id=project_id),
+                project_entity=project_entity,
             )
             linkable_types = schema["entity"]["properties"]["valid_types"]["value"]
             type_whitelist = type_whitelist.union(set(linkable_types))
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -203,7 +203,7 @@ class ShotgunAPI(object):
         """
         # First, let's see if this is even an entity type that we need to worry
         # about. If it isn't, we can just tell the client to stop waiting.
-        if data["entity_type"] not in self._get_entity_type_whitelist():
+        if data["entity_type"] not in self._get_entity_type_whitelist(data.get("project_id")):
             logger.debug(
                 "Entity type %s is not supported, no actions will be returned.",
                 data["entity_type"],
@@ -293,6 +293,7 @@ class ShotgunAPI(object):
                     entity["id"],
                 )
                 pc_data["lookup_hash"] = lookup_hash
+                pc_data["descriptor"] = pc_descriptor
                 contents_hash = pc_data["contents_hash"]
                 cached_data = []
 
@@ -431,7 +432,7 @@ class ShotgunAPI(object):
 
         :param dict data: The data passed down from the wss client.
         :param dict config_data: A dictionary that contains, at a minimum,
-            "lookup_hash", "contents_hash", and "entity" keys.
+            "lookup_hash", "contents_hash", "descriptor", and "entity" keys.
         """
         logger.debug("Caching engine commands...")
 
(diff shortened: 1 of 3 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -650,6 +669,31 @@ def _get_entities_from_payload(self, data):
...
+        Gets a set of entity types that are supported by the browser
+        integration. This set is built from a list of constant entity
+        types, plus all entity types that a PublishedFile entity is
+        allowed to link to, per the current site's schema.
+
+        :returns: A set of string entity types.
+        :rtype: set
+        """
+        if self.ENTITY_TYPE_WHITELIST not in self._cache:
+            type_whitelist = constants.BASE_ENTITY_TYPE_WHITELIST
```
jfboismenu: Why is the base entity type whitelist required? Wouldn't the schema_field_read return those entities as well?

jfboismenu: I know we have fields that are custom for certain projects. Do you know if the `PublishedFile.entity` field can have different list of entities it can refer to between projects? Maybe @nemoDreamer or @francoisjacques could chime in?

francoisjacques: TL;DR - There's a common set for the whole site, and some entities can be masked on a per-project basis.

Boring details:

The list of possibles entity types to a given field is determined by the data_type_properties property of the properties field of a DisplayColumn. (still with me?)

```
DisplayColumn
   - properties
      - data_type_properties
         - allowed_entity_types
```

(there's a special case where the allowed_entity_types isn't set on the DisplayColumn instance itself but determined by the field's Rails association, but I digress)

The display columns are configured at the entity level (site-wide) not at a site level. 

Now,  there are UIVisibilityRules which can be set that makes a given entity type visible or not in a given project. 

Invoking schema_field_read takes into account both the site-wide allowed entity types _and_ the per-project UIVisibilityRules. 

For reference:

https://github.com/shotgunsoftware/shotgun/blob/develop/7.1/app/apis/api3_handler.rb#L1272
https://github.com/shotgunsoftware/shotgun/blob/develop/7.1/app/apis/api3_handler.rb#L3511
https://github.com/shotgunsoftware/shotgun/blob/develop/7.1/app/models/display_column.rb#L534

francoisjacques: @denispasin FYI ^^

jfboismenu: Thanks @francoisjacques !
@thebeeland Seems like you will have to take this into account. You'll need a per-project cache of these entities.

nemoDreamer: 👍 &times; 💯 to @francoisjacques for that answer...! I wouldn't have come close to that...

jfboismenu:
NOTE: This should be merged in master now. The develop/zero_config branch is now dead.

nemoDreamer:
@jfboismenu are we already merging to `master`? or did you mean `develop/7.1`?

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Thanks, guys!

thebeeland:
@nemoDreamer He's referring to tk-framework-desktopserver and not the web app when he talks about mergining to master.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -218,6 +227,7 @@ def cache(cache_file, data, base_configuration, engine_name, config_data):
         arg_data["base_configuration"],
         arg_data["engine_name"],
         arg_data["config_data"],
+        arg_data["config_is_mutable"],
```
jfboismenu: I would rather abbreviate both instances of the word configuration or not abbreviate both here.
`base_configuration`/`configuration_is_mutable` or `base_config` and `config_is_mutable`

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -650,6 +672,48 @@ def _get_entities_from_payload(self, data):
...
+            project_entity = None
+        else:
+            project_entity = dict(type="Project", id=project_id)
+            logger.debug("Looking up schema for project %s", project_entity)
+
+        cache_is_initialized = self.ENTITY_TYPE_WHITELIST in self._cache
+        project_in_cache = project_id in self._cache.get(self.ENTITY_TYPE_WHITELIST, dict())
+
+        if not cache_is_initialized or not project_in_cache:
+            type_whitelist = constants.BASE_ENTITY_TYPE_WHITELIST
```
jfboismenu: I think one of my questions got overshadowed by the discussion about per-project caching on the previous review. Why do you need a base entity type whitelist here? Why wouldn't a call to schema_field_read be enough here? Wouldn't these entities be returned by the call anyway? If it is still required, it would be appreciated if you could provide a comment next to where the constant is defined explaining why this list is required.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -71,6 +71,8 @@ def cache(cache_file, data, base_configuration, engine_name, config_data):
         dict is keyed by pipeline config entity id, each containing a dict
         that contains, at a minimum, "entity", "lookup_hash", and
         "contents_hash" keys.
+    :param bool config_is_mutable: Whether the pipeline config is mutable. If
```
jfboismenu: Same comment as below for the argument `config_is_mutable`.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: It's still required. The PublishedFile's entity link types does not typically include the types in the constant, but we always want to support those, regardless.

PR #29, opened 2017-05-18
Adds more logging and fixes a Core swapping issue

jfboismenu:


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -145,7 +145,7 @@ def execute_action(self, data):
                 name=data["name"],
                 entities=entities,
                 project=project_entity,
-                sys_path=sys.path,
+                sys_path=self._compute_sys_path(),
```
manneohrstrom: a small comment here explaining why we cannot just grab it i think could be useful.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -463,6 +463,16 @@ def _db_connect(self):
             with contextlib.closing(connection.cursor()) as cursor:
                 yield (connection, cursor)
 
+    def _compute_sys_path(self):
```
manneohrstrom: could this be something that we added to the toolkit manager instead (as a classmethod or static i guess, since it's kind of independent of the state of any instance). THinking about sphinx docs - could be nice to include it as part of manager - if nothing else so just for completeness, a dev reading the tk manager docs would be aware of the fact that you cannot grab the sys.path of core due to the magic of the swaps.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -463,6 +463,16 @@ def _db_connect(self):
...
+        """
+        :returns: Path to the current core.
+        """
+        # While core swapping, the Python path is not updated with the new core's Python path,
+        # so make sure the current core is at the front of the Python path for out subprocesses.
+        tank_folder = os.path.dirname(sgtk.__file__)
```
manneohrstrom: debug logging here to include the path that we are adding to sys.path could be helpful for debugging!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -39,6 +39,8 @@ def bootstrap(data, base_configuration, engine_name, config_data):
     # The local import of sgtk ensures that it occurs after sys.path is set
     # to what the server sent over.
     import sgtk
+    sgtk.LogManager().initialize_base_file_handler("tk-shotgun")
```
manneohrstrom: yes please! i think i commented on this in previous CRs. Great to see it added.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: I left that comment inside of the method instead of pasting it twice.

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: It's actually dependent on the state of the sgtk instance. This is not something the manager can give us.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -463,6 +463,16 @@ def _db_connect(self):
             with contextlib.closing(connection.cursor()) as cursor:
                 yield (connection, cursor)
 
+    def _compute_sys_path(self):
+        """
+        :returns: Path to the current core.
+        """
+        # While core swapping, the Python path is not updated with the new core's Python path,
+        # so make sure the current core is at the front of the Python path for out subprocesses.
+        python_folder = sgtk.bootstrap.ToolkitManager.get_core_python_path()
```
manneohrstrom: ❤️

PR #30, opened 2017-05-19
For #42750, #42771, #42772: QA fixes.

thebeeland:
* Fixes a bug that caused failed command executions to not report any output back to the client.
* Implements a custom log handler in `execute_command.py` that outputs to stdout and tags all log messages in such a way that...
* ...stdout can be filtered, cleaned up, and reported back to the client when a command successfully completes.
* Takes ownership of the `browser_integration.py` hook.
* Removes HTML output from core and app upgrade commands, replacing it with simple markdown.

### DIFF
--- a/hooks/browser_integration.py
+++ b/hooks/browser_integration.py
@@ -1,0 +1,102 @@
+# Copyright (c) 2017 Shotgun Software Inc.
+# 
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Hook that provides utilities used by Toolkit's browser integration.
+"""
+
+import sgtk
+
+HookBaseClass = sgtk.get_hook_baseclass()
+
+class BrowserIntegration(HookBaseClass):
+    def get_cache_key(self, config_uri, project, entity_type):
+        """
+        Computes the key used to uniquely identify a row in the cache database for
+        the given pipeline configuration uri and entity type. This method can be used
+        to change how a specific entity type is cached, should that be desired.
+
+        .. Note:
+            Changing the logic in this method will invalidate related cache entries
+            that already exist in the cache.
+
+        :param str config_uri: A descriptor uri for the pipeline configuration.
+        :param dict project: The project entity.
+        :param str entity_type: The entity type.
+
+        :returns: A uniquely-identifiable key.
+        :rtype: str
+        """
+        return "%s@%s" % (config_uri, entity_type)
+
+    def get_site_state_data(self):
+        """
+        Builds a list of dictionaries matching the kwargs expected by the
+        Shotgun Python API's find method. These will be queried at the time
+        when a contents state hash is generated during cache population or
+        lookup. This hash is then used to verify that the data found in the
+        cache is still valid. Customization of this method's logic can be
+        used to control what determines whether cached data is valid or not.
+        By default, we track the state of the Shotgun site's Software entities
+        to determine whether any fields there have changed since the data was
+        cached.
+
+        .. Note:
+            Changing the logic in this method will invalidate related cache
+            entries that already exist in the cache.
+
+        :returns: A list of dictionaries containing kwargs that will be passed
+            to the Shotgun Python API's find method.
+        :rtype: list
+        """
+        data_requests = []
+        data_requests.append(
+            dict(
+                entity_type="Software",
+                filters=[],
+                fields=self.parent.shotgun.schema_field_read("Software").keys(),
+            )
+        )
+
+        return data_requests
+
+    def process_commands(self, commands, project, entities):
+        """
+        Allows for pre-processing of the commands that will be returned to the client.
+        Overriding the logic in this method could be used, for example, to filter out
+        certain commands from being passed up to the client.
+
+        :param list commands: A list of dictionaries, each containing information
+            about a command to be passed up to the client. Each dict will include
+            the following keys: name, title, app_name, deny_permissions, and
+            supports_multiple_selection.
+        :param dict project: The project entity.
+        :param list entities: A list of entity dictionaries representing all entities
+            that were selected in the web client.
+
+        .. Example:
+            dict(
+                name="launch_maya",
+                title="Maya 2017",
+                deny_permissions=[],
+                supports_multiple_selection=False,
+                app_name="tk-multi-launchapp",
+                group="Maya",
+                group_default=True,
+                engine_name="tk-maya",
+            )
+
+        :returns: Processed commands.
+        :rtype: list
+        """
+        return commands
+
+
+
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -787,8 +787,8 @@ class ShotgunAPI(object):
             entity_type = self._get_task_parent_entity_type(entity_id)
             logger.debug("Task entity's parent entity type: %s", entity_type)
 
-        return self._engine.sgtk.execute_core_hook_method(
-            "browser_integration",
+        return self._bundle.execute_hook_method(
+            "browser_integration_hook",
             "get_cache_key",
             config_uri=config_uri,
             project=project,
@@ -934,7 +934,7 @@ class ShotgunAPI(object):
     def _get_site_state_data(self):
         """
         Gets state-related data for the site. Exactly what data this is depends
-        on the "browser_integration" core hook's "get_site_state_data" method,
+        on the "browser_integration" hook's "get_site_state_data" method,
         which returns a list of dicts passed to the Shotgun Python API's find
         method as kwargs. The data returned by this method is cached based on
         the WSS connection key provided to the API's constructor at instantiation
@@ -947,8 +947,8 @@ class ShotgunAPI(object):
         if self.SITE_STATE_DATA not in self._cache:
             self._cache[self.SITE_STATE_DATA] = []
 
-            requested_data_specs = self._engine.sgtk.execute_core_hook_method(
-                "browser_integration",
+            requested_data_specs = self._bundle.execute_hook_method(
+                "browser_integration_hook",
                 "get_site_state_data",
             )
 
@@ -1020,7 +1020,7 @@ class ShotgunAPI(object):
         """
         Filters out commands that are not associated with an app, and then
         calls the process_commands methods from the browser_integration
-        core hook, returning the result.
+        hook, returning the result.
 
         :param list commands: The list of command dictionaries to be processed.
         :param dict project: The project entity.
@@ -1043,8 +1043,8 @@ class ShotgunAPI(object):
                     "Command %s filtered out for browser integration.", command
                 )
 
-        return self._engine.sgtk.execute_core_hook_method(
-            "browser_integration",
+        return self._bundle.execute_hook_method(
+            "browser_integration_hook",
             "process_commands",
             commands=filtered,
             project=project,
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -21,6 +21,7 @@ import json
 import fnmatch
 import datetime
 import copy
+import base64
 
 import sgtk
 from sgtk.commands.clone_configuration import clone_pipeline_configuration_html
@@ -200,15 +201,28 @@ class ShotgunAPI(object):
         # handler that the execute_command script builds, and we
         # remove that header from those lines and keep them so that
         # they're passed up to the client.
-        for line in stdout.split("\n"):
-            if line.startswith("SGTK_LOG_OUTPUT:"):
-                filtered_output.append(re.sub(r"^SGTK_LOG_OUTPUT:\s*", "", line))
+        tag = "SGTK:"
+        tag_length = len(tag)
+
+        # We check both stdout and stderr. We identify lines that start with
+        # our tag, and if we find one, we remove the tag, and then base64
+        # decode the rest of the message. The message is encoded this way
+        # because it collapses multi-line log messages into a single line
+        # of text. This is important, because there's only one tag per log
+        # message, and we would otherwise filter out everything in the log
+        # message that wasn't on the first line of text before any newlines.
+        for line in stdout.split("\n") + stderr.split("\n"):
+            if line.startswith(tag):
+                filtered_output.append(base64.b64decode(line[tag_length:]))
+
+        filtered_output_string = "\n".join(filtered_output)
+        logger.debug("Filtered and decoded log output: %s", filtered_output_string)
 
         logger.debug("Command execution complete.")
         self.host.reply(
             dict(
                 retcode=constants.COMMAND_SUCCEEDED,
-                out="\n".join(filtered_output),
+                out=filtered_output_string,
                 err="",
             ),
         )
@@ -769,6 +783,10 @@ class ShotgunAPI(object):
 
         if not cache_is_initialized or not project_in_cache:
             type_whitelist = constants.BASE_ENTITY_TYPE_WHITELIST
+
+            # This will only ever happen once per unique connection. That means
+            # on page refresh it happens, but not every time menu actions are
+            # requested.
             schema = self._engine.shotgun.schema_field_read(
                 constants.PUBLISHED_FILE_ENTITY,
                 field_name="entity",
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -150,6 +150,7 @@ class ShotgunAPI(object):
                 sys_path=self._compute_sys_path(),
                 base_configuration=constants.BASE_CONFIG_URI,
                 engine_name=constants.ENGINE_NAME,
+                logging_prefix=constants.LOGGING_PREFIX,
             ),
         )
 
@@ -201,7 +202,7 @@ class ShotgunAPI(object):
         # handler that the execute_command script builds, and we
         # remove that header from those lines and keep them so that
         # they're passed up to the client.
-        tag = "SGTK:"
+        tag = constants.LOGGING_PREFIX
         tag_length = len(tag)
 
         # We check both stdout and stderr. We identify lines that start with
@@ -216,9 +217,8 @@ class ShotgunAPI(object):
                 filtered_output.append(base64.b64decode(line[tag_length:]))
 
         filtered_output_string = "\n".join(filtered_output)
-        logger.debug("Filtered and decoded log output: %s", filtered_output_string)
-
         logger.debug("Command execution complete.")
+
         self.host.reply(
             dict(
                 retcode=constants.COMMAND_SUCCEEDED,
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -10,6 +10,7 @@
 
 import sys
 import os
+import re
 import cPickle
 import sqlite3
 import json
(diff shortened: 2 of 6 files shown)


python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
     from sgtk.commands.core_upgrade import TankCoreUpdater
-    logger = sgtk.LogManager.get_logger(LOGGER_NAME)
-
-    code_css_block = "display: block; padding: 0.5em 1em; border: 1px solid #bebab0; background: #faf8f0;"
 
     # Create an upgrader instance that we can query if the install is up to date.
     installer = TankCoreUpdater(
         engine.sgtk.pipeline_configuration.get_install_location(),
-        logger,
+        engine._log,
```
thebeeland: Thoughts on this, @manneohrstrom ?

thebeeland: Maybe `engine.sgtk.log` instead? Though that's not really public, either.

hooks/browser_integration.py
```diff
@@ -0,0 +1,102 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Hook that provides utilities used by Toolkit's browser integration.
```
manneohrstrom: this is so awesome

hooks/browser_integration.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+
+import sgtk
+
+HookBaseClass = sgtk.get_hook_baseclass()
+
+class BrowserIntegration(HookBaseClass):
+    def get_cache_key(self, config_uri, project, entity_type):
```
manneohrstrom: i think i might have asked this in a previous PR - how difficult would it be to pass down a full sg data dict instead of project and entity_type? I am thinking about future use cases where we for example want separate menus for shot tasks and asset tasks. in that case, i am guessing we nneed to key the hash off both entity type , config uri and the entity type of the associated entity?

hooks/browser_integration.py
```diff
@@ -0,0 +1,102 @@
...
+        :returns: A list of dictionaries containing kwargs that will be passed
+            to the Shotgun Python API's find method.
+        :rtype: list
+        """
+        data_requests = []
+        data_requests.append(
+            dict(
+                entity_type="Software",
+                filters=[],
+                fields=self.parent.shotgun.schema_field_read("Software").keys(),
```
manneohrstrom: how often is this hook called? `schema_field_read` is an expensive operation, so might be worth adding in the docstring how frequently this is called to give implementors an idea what kind of stuff can be added here.

manneohrstrom: should we be worried about sites earlier than 7.2? similar to the comment above, maybe add some docstring notes to explain when this is being called.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -183,14 +184,31 @@ def execute_action(self, data):
...
+        filtered_output = []
+
+        # We need to filter stdout before we send it to the client.
+        # We look for lines that we know came from the custom log
+        # handler that the execute_command script builds, and we
+        # remove that header from those lines and keep them so that
+        # they're passed up to the client.
+        for line in stdout.split("\n"):
+            if line.startswith("SGTK_LOG_OUTPUT:"):
+                filtered_output.append(re.sub(r"^SGTK_LOG_OUTPUT:\s*", "", line))
```
manneohrstrom: nitpick: this feels like a very expensive way to filter. Why not just a string manipulation, e.g.

```
>>> x = "TOOLKIT:foo bar"
>>> x[len("TOOLKIT:"):]
'foo bar'
```

Also, would be great to make `SGTK_LOG_OUTPUT` a constant.

manneohrstrom: what happens to stderr? We should filter it, even in the case of success!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -787,8 +805,8 @@ def _get_lookup_hash(self, config_uri, project, entity_type, entity_id):
             entity_type = self._get_task_parent_entity_type(entity_id)
             logger.debug("Task entity's parent entity type: %s", entity_type)
 
-        return self._engine.sgtk.execute_core_hook_method(
-            "browser_integration",
+        return self._bundle.execute_hook_method(
```
manneohrstrom: ❤️

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -85,26 +85,22 @@ def cache(cache_file, data, base_configuration, engine_name, config_data, config
         it is, then we include the __core_info and __upgrade_check commands.
     """
     engine = bootstrap(data, base_configuration, engine_name, config_data)
-
-    import sgtk
-    logger = sgtk.LogManager.get_logger(LOGGER_NAME)
-
-    logger.debug("Raw payload from client: %s", data)
+    engine.log_debug("Raw payload from client: %s" % data)
```
manneohrstrom: can we add a comment here to help future code maintainers - just a note to say that the engine returned here may be linked to an older core etc.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -29,26 +29,19 @@ def app_upgrade_info(engine):
 
     :param engine: The currently-running engine instance.
     """
-    import sgtk
-    logger = sgtk.LogManager.get_logger(LOGGER_NAME)
-
-    code_css_block = "display: block; padding: 0.5em 1em; border: 1px solid #bebab0; background: #faf8f0;"
-
-    logger.info(
+    engine.log_info(
```
manneohrstrom: please add a comment that this uses markdown and this is safe because this code will only ever be executed on sg sites with 7.2+

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
         "You are currently running version %s of the Shotgun Pipeline Toolkit." % cv
     )
 
     if not engine.sgtk.pipeline_configuration.is_localized():
-        logger.info("")
-        logger.info(
-            "Your core API is located in <code>%s</code> and is shared with other "
+        engine.log_info(
```
manneohrstrom: same comment about markdown here would be helpful i think.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
manneohrstrom: interesting. firstly, definitely add a comment explaining that we cannot use `engine.logger` because we have to support back to 0.16 cores. yes, i agree `engine.sgtk.log` is the way to go - but again, please add a comment explaining that this `sgtk.log` property is a special thing that only exists for the shotgun engine etc.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
+    The pre-engine-start callback that's given to the bootstrap API.
+    This callback handles attaching a custom logger to SGTK prior to
+    the Shotgun engine being initialized. This allows us to customize
+    the output of the logger in such a way that it is easily identified
+    and filtered before going back to the client for display.
+
+    :param context: The context object being used during the bootstrap
+        process.
+    """
+    import sgtk
```
manneohrstrom: what core is this? guaranteed to be 0.18? or the core that we are bootstrapping into? please add comment

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
+
+    sgtk.LogManager().initialize_base_file_handler("tk-shotgun")
```
manneohrstrom: this is probably in the wrong place - please note that @jfboismenu added it last night as part of his changes: https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/master/python/tk_framework_desktopserver/shotgun/scripts/execute_command.py#L140

So we don't need it here (and i am worried that `sgtk` can be a 0.17 core at this point).

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
+
+    # We need to make sure messages from this logger end up going to
+    # stdout. We'll be trapping stdout from the RPC API, which will
+    # give us the output that gets sent back to the client when the
+    # command is completed.
+    handler = logging.StreamHandler(sys.stdout)
+    sgtk.LogManager().initialize_custom_handler(handler)
```
manneohrstrom: again, do we know we have the LogManager()? 

I was looking at this code last night (just before i handed over to you!), and i was thinking about a slightly different approach - curious what you think:

- in the `bootstrap()` method, call the `LogManager` get a good logging root for the sg engine. See [here](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L114): `logger = LogManager.get_logger("env.%s.%s" % (env.name, engine_instance_name))` it's a bit hacky, but if we want to be backwards compatible it was the best thing i could think of. Can you think of anything better? That would give the `tk-shotgun` engine a standard logger writing to the correct place in the hierarchy. 

- Still in the `bootstrap()` metod - before we call the sgtk bootstrap, set up the log handler that prints to stdout. This is the place where we know exactly what core we are running so it's easier to write safe code.

- as we set up the params for bootstrappig, we make sure pass the logger we created above to the `pre_engine_start_callback` 

- `pre_engine_start_callback` just does one thing - hooks up the logger with `sgtk.log` - since we don't know what core runs in the `pre_engine_start_callback` we want to keep the footprint inside here as lightweight as possible.

This means logging starts earlier so we can track say errors at engine startup. It also has a minimal footprint in old-core territory.

what do you think?

python/tk_framework_desktopserver/shotgun/api_v2.py
jfboismenu: Agreed. Especially since you'll be eating all the extra whitespace which may have been on purpose.

jfboismenu: Same concerns about stderr.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
jfboismenu: Yeah, using sgtk.log here is fine because we know it exists since where the ones who put it there in the first place during `pre_engine_start_callback`.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
         engine.sgtk.pipeline_configuration.get_install_location(),
-        logger,
+        engine._log,
     )
 
     cv = installer.get_current_version_number()
     lv = installer.get_update_version_number()
 
-    logger.info(
+    engine.log_info(
```
jfboismenu: Did you run the code? I see the same errors I saw in my original branch. The `install_root` variable is never set and the code will crash. Also, you've switched everything to italics instead of bold like it used to be.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
jfboismenu: This code is not 0.18. This hook is run after core swapping.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +52,87 @@ def core_info(engine):
...
+    # We need to make sure messages from this logger end up going to
+    # stdout. We'll be trapping stdout from the RPC API, which will
+    # give us the output that gets sent back to the client when the
+    # command is completed.
+    handler = logging.StreamHandler(sys.stdout)
+    sgtk.LogManager().initialize_custom_handler(handler)
+
+    # Give it an easily-identifiable format. We'll use this in the RPC API
+    # when filtering stdout before passing it up to the client.
+    formatter = logging.Formatter("SGTK_LOG_OUTPUT: %(message)s")
```
jfboismenu: If a log method is called with a multi-line message, this will be in `stdout`:

```
SGTK_LOG_OUTPUT: This is my
multiline message
```
This means we will only send back `This is my` to the browser.

You need to write your own formatter that will cgi.escape the output or base64 encode it so that on the other end you unescape or unbase64 it before sending it back to the browser.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
jfboismenu: For reference, here's what I had started with: https://github.com/shotgunsoftware/tk-framework-desktopserver/commit/2a1b5d1667a6176b2c36a0c611d1576c774a69b3

hooks/browser_integration.py
thebeeland: We handle shot vs. asset tasks already. The data required to make that determination is not passed down from SG, however; it's queried and cached (both in the web app and in Python).

hooks/browser_integration.py
thebeeland: It's called once per connection and cached in memory after that. It'll be requeried on page refresh, but only once.

thebeeland: No, the v2 RPC API won't be used by pre 7.2 Shotgun. Anything older is routed to v1.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
thebeeland: Thanks for the link.

As for bold vs. italics, *foo* shows as bold in SG after being run through the markup processor Phil had you use in the web app when handling returned output from the server.

jfboismenu: Really. I'm surprised...
`*foo*`: *foo*
`**foo**`: **foo**

thebeeland: Will chat with Phil on why that's the case. It didn't strike me as odd last night, but it does now that JF brings it up.

thebeeland: The answer is that we follow Slack-style markdown, which uses `*foo*` for bold, and `_foo_` for italics. Unfortunately, the world is full of variations on markdown. :|

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
thebeeland: I added comments before the first log_info call for each of the two functions here that make use of markdown.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -183,14 +185,44 @@ def execute_action(self, data):
...
             return
 
+        filtered_output = []
+
+        # We need to filter stdout before we send it to the client.
+        # We look for lines that we know came from the custom log
+        # handler that the execute_command script builds, and we
+        # remove that header from those lines and keep them so that
+        # they're passed up to the client.
+        tag = "SGTK:"
```
manneohrstrom: why not in `constants`?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -183,14 +185,44 @@ def execute_action(self, data):
...
+        # because it collapses multi-line log messages into a single line
+        # of text. This is important, because there's only one tag per log
+        # message, and we would otherwise filter out everything in the log
+        # message that wasn't on the first line of text before any newlines.
+        for line in stdout.split("\n") + stderr.split("\n"):
+            if line.startswith(tag):
+                filtered_output.append(base64.b64decode(line[tag_length:]))
+
+        filtered_output_string = "\n".join(filtered_output)
+        logger.debug("Filtered and decoded log output: %s", filtered_output_string)
```
manneohrstrom: do we need to log all the output like this or is this a bit too much potentially?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -127,12 +128,12 @@ def cache(cache_file, data, base_configuration, engine_name, config_data, config
         ])
     else:
         if config_is_mutable:
-            logger.debug("Not a Project entity: not registering core and app update commands.")
+            engine.log_debug("Not a Project entity: not registering core and app update commands.")
```
manneohrstrom: it seems a little odd that we are testing for `if config_is_mutable` but then the debug message says "not a project entity:"

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -12,15 +12,30 @@
...
+        """
+        Encodes log messages as base64. This allows us to collapse multiline
+        log messages into a single line of text. Before presenting the log
+        messages to a user, the caller of execute_command.py will be required
+        to decode the message. Every message is tag at its head with "SGTK:",
+        making output from a logger using this formatter easily identifiable.
+        """
+        result = super(_Formatter, self).format(*args, **kwargs)
+        return "{tag}:{log_data}".format(
+            tag="SGTK",
```
manneohrstrom: i suggest pick this up from constants

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -12,15 +12,30 @@
...
+    def format(self, *args, **kwargs):
+        """
+        Encodes log messages as base64. This allows us to collapse multiline
+        log messages into a single line of text. Before presenting the log
+        messages to a user, the caller of execute_command.py will be required
+        to decode the message. Every message is tag at its head with "SGTK:",
+        making output from a logger using this formatter easily identifiable.
+        """
+        result = super(_Formatter, self).format(*args, **kwargs)
+        return "{tag}:{log_data}".format(
```
manneohrstrom: minor/nitpick: why use format here when we use % style formatting everywhere else? seems inconsequential.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -59,67 +71,83 @@ def core_info(engine):
...
+    This callback handles attaching a custom logger to SGTK prior to
+    the Shotgun engine being initialized. This allows us to customize
+    the output of the logger in such a way that it is easily identified
+    and filtered before going back to the client for display.
+
+    :param logger: The logger to inject into the sgtk api instance.
+    :param context: The context object being used during the bootstrap
+        process.
+    """
+    context.sgtk.log = logger
```
manneohrstrom: nice!!!

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Good point!

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Maybe a bit much since it'll all be in tk-shotgun.log. I had put that in there while debugging just so I had a single log file to tail, but I think it's worth removing this for release.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
thebeeland: Thinking about this, I'm not sure it's going to be easy to get an import of constants out of the framework given that this is running as a subprocess. I'll take a quick crack at it, but won't spend too much time trying to make it work.

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -154,6 +203,14 @@ def bootstrap(config, base_configuration, entity, engine_name):
     manager.plugin_id = "basic.shotgun"
     manager.base_configuration = base_configuration
 
+    # By building a partial object, we can go ahead and attach the logger
+    # to the callback function, where it will become the first argument
+    # at call time.
+    manager.pre_engine_start_callback = functools.partial(
```
jfboismenu: +1 for using partial instead of a global! :D

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -303,6 +301,9 @@ def execute(config, project, name, entities, base_configuration, engine_name):
 
     sys.path = arg_data["sys_path"]
 
+    global LOGGING_PREFIX
```
jfboismenu: I think your global isn't required, as you are already in the global scope.

PR #33, opened 2017-06-03
For #43057: Implements a legacy code path for classic sgtk projects.

thebeeland:
If a project contains classic-style PipelineConfiguration entities, then its get_actions and engine command execution routines call through a legacy code path. This legacy path makes use of the tank command, as the v1 wss API did.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -78,6 +78,13 @@ class ShotgunAPI(object):
         self._logger = sgtk.platform.get_logger("api-v2")
         self._process_manager = process_manager
 
+        if constants.DISABLE_LEGACY_BROWSER_INTEGRATION_WORKAROUND in os.environ:
+            logger.debug("Legacy tank command pathway disabled.")
+            self._allow_legacy_workaround = False
+        else:
+            logger.debug("Legacy tank command pathway allowed for classic configs.")
+            self._allow_legacy_workaround = True
+
         if self._wss_key not in self.WSS_KEY_CACHE:
             self.WSS_KEY_CACHE[self._wss_key] = dict()
 
@@ -157,12 +164,15 @@ class ShotgunAPI(object):
         # get_actions method ran through its own legacy code path. In that case, we
         # shell out to the tank command by way of our process_manager object instead
         # of using the modern code path.
-        if command_name not in ["__core_info", "__upgrade_check"]:
-            if "_legacy_config_root" in config_entity:
+        if command_name not in constants.LEGACY_EXEMPT_ACTIONS and self._allow_legacy_workaround:
+            if constants.LEGACY_CONFIG_ROOT in config_entity:
+                # The arguments list is the name of the command, then the entity
+                # type, and then a comma-separated list of entity ids.
+                entity_ids = [str(e["id"]) for e in entities]
                 (out, err, retcode) = self.process_manager.execute_toolkit_command(
-                    config_entity["_legacy_config_root"],
+                    config_entity[constants.LEGACY_CONFIG_ROOT],
                     "shotgun_run_action",
-                    [data["name"], entities[0]["type"], str(entities[0]["id"])],
+                    [data["name"], entities[0]["type"], ",".join(entity_ids)],
                 )
 
                 # Sanitize the output. By going the legacy route here, we're going
@@ -171,13 +181,11 @@ class ShotgunAPI(object):
                 # is going to believe that we're sending v2-style output, which is
                 # taken and displayed as is, and is assumed to be markdown and not
                 # HTML.
-                out = self._legacy_sanitize_output(out)
-
                 self.host.reply(
                     dict(
                         retcode=retcode,
-                        err=err,
-                        out=out,
+                        err=self._legacy_sanitize_output(err),
+                        out=self._legacy_sanitize_output(out),
                     )
                 )
 
@@ -324,7 +332,6 @@ class ShotgunAPI(object):
         manager.base_configuration = constants.BASE_CONFIG_URI
         manager.bundle_cache_fallback_paths = self._engine.sgtk.bundle_cache_fallback_paths
 
-        all_actions = dict()
         all_pc_data = self._get_pipeline_configuration_data(
             manager,
             project_entity,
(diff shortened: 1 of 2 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1079,6 +1153,134 @@ def _get_task_parent_entity_type(self, task_id):
...
+        project_actions = self._legacy_get_project_actions(config_paths)
+        all_actions = dict()
+        config_names = []
+
+        for config_name, config_data in config_data.iteritems():
+            config_path, config_entity = config_data
+            commands = []
+            get_actions_data = project_actions[config_path]["shotgun_get_actions"]
+            env_file_name = "shotgun_%s.yml" % entity_type.lower()
+            raw_actions_data = get_actions_data[env_file_name]
```
thebeeland: NOTE: This isn't correct. It needs to be a .get and then report an unsupported entity type to the client if None. I'll fix it Monday.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1079,6 +1153,134 @@ def _get_task_parent_entity_type(self, task_id):
...
+
+            if raw_actions_data["retcode"] != 0:
```
thebeeland: This also needs to change, most likely. We shouldn't just skip forward without reporting the error to the client.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -140,10 +141,51 @@ def execute_action(self, data):
...
+        # of these "special" project-level commands already outputs markdown syntax,
+        # which is preferable to the HTML we would get from the tank command.
+        #
+        # The second bit is that we can identify commands that need to be run in
+        # legacy mode by whether the config entity dictionary we got from the client
+        # contains a "_legacy_config_root" key that was shoved into it when the
+        # get_actions method ran through its own legacy code path. In that case, we
+        # shell out to the tank command by way of our process_manager object instead
+        # of using the modern code path.
+        if command_name not in ["__core_info", "__upgrade_check"]:
```
jfboismenu: Nitpick: you could collapse those two `ifs` into a single if, unless you feel its less readable.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -305,6 +331,54 @@ def get_actions(self, data):
...
+        legacy_config_data = dict()
+
+        # A classic-style config will contain a roots.yml file that defines
+        # a set of storage roots. We can access that by way of the descriptor
+        # object that we already have. If there are required storages, then
+        # we know we're in a classic situation.
+        for config in config_entities:
+            if config["descriptor"].required_storages:
+                legacy_config_data[config["name"]] = (
+                    config["descriptor"].get_path(),
```
jfboismenu: If [this comment](https://github.com/shotgunsoftware/tk-core/pull/469#pullrequestreview-42064764) is correct, it means that this line is not valid anymore as well as any other places where you call `get_path` either.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -140,10 +141,51 @@ def execute_action(self, data):
...
+        #
+        # The second bit is that we can identify commands that need to be run in
+        # legacy mode by whether the config entity dictionary we got from the client
+        # contains a "_legacy_config_root" key that was shoved into it when the
+        # get_actions method ran through its own legacy code path. In that case, we
+        # shell out to the tank command by way of our process_manager object instead
+        # of using the modern code path.
+        if command_name not in ["__core_info", "__upgrade_check"]:
+            if "_legacy_config_root" in config_entity:
+                (out, err, retcode) = self.process_manager.execute_toolkit_command(
```
jfboismenu: If I understand correctly this means that we're now a bit faster than v1. This looks like we'll now only get actions for entities that the browser asks us for instead of simply scanning every shotgun_*.yml files like before. 

Correct?

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Understood. That change in tk-core made me nervous, anyway. We'll find another route to take.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: It should be faster than v1, but I don't really have a sense of how much. Regardless, it's certainly not going to be slower, which is nice.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -305,6 +331,54 @@ def get_actions(self, data):
...
+        # The first thing we do is check to see if we're dealing with a
+        # classic SGTK setup. In that case, we're going to short-circuit
+        # the get_actions call and go into a legacy setup that makes use
+        # of the "tank" command by way of this api's process_manager.
+        config_entities = [d["entity"] for i, d in all_pc_data.iteritems()]
+        legacy_config_data = dict()
+
+        # A classic-style config will contain a roots.yml file that defines
+        # a set of storage roots. We can access that by way of the descriptor
+        # object that we already have. If there are required storages, then
```
manneohrstrom: this roots.yml check isn't a good way to check if a config is classic or not. @jfboismenu @thebeeland maybe the `resolve_descriptor()` method on bootstrap should instead return something like a `Configuration` object which contains methods and properties to resolve descriptor, indicate if a config is classic etc? Inside bootstrap we have a `Configuration` object hierarchy to expose that - maybe we should consider exposing that? That would require some discussion...

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1079,6 +1153,134 @@ def _get_task_parent_entity_type(self, task_id):
...
+        legacy v1 API method of getting toolkit actions and the v2 menu
+        logic in the Shotgun web app versions 7.2.0+.
+
+        :param dict config_data: A dictionary, keyed by PipelineConfiguration
+            entity name, containing a tuple of config root path and config
+            entity dict, in that order.
+        :param str entity_type: The entity type that we're getting actions
+            for.
+        """
+        config_paths = [p[0] for n, p in config_data.iteritems()]
```
manneohrstrom: please comment this `[0]` index

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1079,6 +1153,134 @@ def _get_task_parent_entity_type(self, task_id):
...
+        tank command. This method filters out any lines of output not wrapped
+        in span tags, and replaces HTML bold tags with Slack-style markdown
+        bold syntax.
+
+        :param str out: The raw output string to sanitize.
+
+        :returns: The sanitized output.
+        :rtype: str
+        """
+        sanitized = []
```
manneohrstrom: `out.replace("<span>", "").replace("</span>", "")`?

Markdown supports html tags, so `<b>` is perfectly valid markdown, so no need to sanitize.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -140,10 +141,51 @@ def execute_action(self, data):
...
+
+                self.host.reply(
+                    dict(
+                        retcode=retcode,
+                        err=err,
+                        out=out,
+                    )
+                )
+
+                return
```
thebeeland: My reply of a semi-nitpick is that I equally dislike HUGE if/else blocks that get the same result as a return. :/

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: I'm concerned about complexity, given that we consider this a short-term workaround. I hesitate to want to redesign anything in core as a result.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Not that I saw in this case. The sanitization I've done was a direct response to what I was seeing in SG. I'll double check, however.

thebeeland: Confirmed that the HTML bold tags aren't supported by SG's markdown.

PR #34, opened 2017-06-07
For #43159, chrome 59 fix

jfboismenu:
Chrome 59 on Mac breaks browser integration because our certificate doesn't report the correct certificate version. Version 3 of the certificate is the one that accepts X509 extensions.

We're updating the framework here to allow an app or engine to regenerate the certificate on demande.

This should be code reviewed in tandem with [this](https://github.com/shotgunsoftware/tk-desktop/pull/76).

### DIFF
--- a/framework.py
+++ b/framework.py
@@ -57,6 +57,8 @@ class DesktopserverFramework(sgtk.platform.Framework):
 
         :param str host: Host for which we desire to answer requests.
         :param int user_id: Id of the user for which we desire to answer requests.
+        :param parent: Parent widget for any pop-ups to show during initialization.
+        :type parent: :class:`PySide.QtGui.QWidget`
         """
         self._tk_framework_desktopserver = self.import_module("tk_framework_desktopserver")
 
@@ -105,6 +107,12 @@ class DesktopserverFramework(sgtk.platform.Framework):
             self.logger.exception("Could not start the browser integration:")
 
     def regenerate_certificates(self, parent=None):
+        """
+        Regenerates the certificates.
+
+        :param parent: Parent widget for any pop-ups to show during certificate generation.
+        :type parent: :class:`PySide.QtGui.QWidget`
+        """
         self.__ensure_certificate_ready(regenerate_certs=True, parent=parent)
 
     def destroy_framework(self):
@@ -120,6 +128,10 @@ class DesktopserverFramework(sgtk.platform.Framework):
         """
         Ensures that the certificates are created and registered. If something is amiss, then the
         certificates are regenerated.
+
+        :param bool regenerate_certs: If ``True``, certificates will be regenerated.
+        :param parent: Parent widget for any pop-ups to show during certificate generation.
+        :type parent: :class:`PySide.QtGui.QWidget`
         """
         cert_handler = self._tk_framework_desktopserver.get_certificate_handler(
             self._settings.certificate_folder
--- a/framework.py
+++ b/framework.py
@@ -47,7 +47,7 @@ class DesktopserverFramework(sgtk.platform.Framework):
 
     ##########################################################################################
     # init and destroy
-    def launch_desktop_server(self, host, user_id):
+    def launch_desktop_server(self, host, user_id, parent=None):
         """
         Initializes the desktop server.
 
@@ -90,7 +90,7 @@ class DesktopserverFramework(sgtk.platform.Framework):
             return
 
         try:
-            self.__ensure_certificate_ready()
+            self.__ensure_certificate_ready(regenerate_certs=False, parent=parent)
 
             self._server = self._tk_framework_desktopserver.Server(
                 keys_path=self._settings.certificate_folder,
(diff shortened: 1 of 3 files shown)


python/tk_framework_desktopserver/certificates.py
```diff
@@ -46,12 +48,41 @@ def exists(self):
...
 
-    def remove_files(self):
+    def backup_files(self):
         """
-        Removes the files from the
+        Backups the certificate files to the backup/year-month-day-hour-minute-second folder.
         """
-        os.unlink(self._cert_path)
-        os.unlink(self._key_path)
+        parent_folder = os.path.dirname(self._cert_path)
```
manneohrstrom: please add debug logging here.

framework.py
```diff
@@ -57,6 +57,8 @@ def launch_desktop_server(self, host, user_id):
 
         :param str host: Host for which we desire to answer requests.
         :param int user_id: Id of the user for which we desire to answer requests.
+        :param parent: Parent widget for any pop-ups to show during initialization.
```
manneohrstrom: instead of passing the parent between methods, could it be an instance variable?

python/tk_framework_desktopserver/certificates.py
```diff
@@ -46,12 +48,41 @@ def exists(self):
...
+        :param dst_folder: Folder in which we will be backing up the file.
+        """
+
+        # If the certificate file does not exist, no need to backup.
+        if not os.path.exists(src_file):
+            return
+
+        # If the backup folder does not exist, create it.
+        if not os.path.exists(dst_folder):
+            os.makedirs(dst_folder)
```
manneohrstrom: better to use https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/filesystem.py#L115 to ensure we write with open permissions?

python/tk_framework_desktopserver/certificates.py
```diff
@@ -46,12 +48,41 @@ def exists(self):
...
+
+    def _backup_file(self, src_file, dst_folder):
+        """
+        Takes a certificate file (if it exists) and moves it to the destination folder.
+
+        :param src_file: File to move.
+        :param dst_folder: Folder in which we will be backing up the file.
+        """
+
+        # If the certificate file does not exist, no need to backup.
```
manneohrstrom: please add debug logging to this method. will make troubleshooting much easier if something weird happens!!!

manneohrstrom: I found in my experience it's good to add lots of debug around all external processes, stuff like I/O permissions, process spawn, etc - sometimes clients have errors which are hard for us to repro and if we have debug logs that thell the story that's super helpful.

python/tk_framework_desktopserver/certificates.py
```diff
@@ -125,7 +160,7 @@ def _check_call(self, ctx, cmd):
 
         # Do not use popen.check_call because it won't redirect stderr to stdout properly
         # and it can't close stdin which causes issues in certain configurations on Windows.
-        logger.debug("%s: %s" % (ctx.capitalize(), cmd))
+        logger.info("%s: %s" % (ctx.capitalize(), cmd))
```
manneohrstrom: can this be abit more informative? e.g. "Created certificate %s...." or something like that?

python/tk_framework_desktopserver/certificates.py
```diff
@@ -226,13 +261,13 @@ def __init__(self, certificate_folder):
 
         # Ensure that the Chrome certificate registry folder exists
         if not os.path.exists(self._PKI_DB_PATH):
-            logger.debug("Creating '%s'", self._PKI_DB_PATH)
+            logger.info("Creating '%s'", self._PKI_DB_PATH)
```
manneohrstrom: `Creating chrome certificate registry folder %s`?

python/tk_framework_desktopserver/certificates.py
```diff
@@ -226,13 +261,13 @@ def __init__(self, certificate_folder):
...
             os.makedirs(self._PKI_DB_PATH)
 
         # If the Chrome certificate registry is empty, create it. If there is already a database in
         # there, the folder won't be empty.
         if not os.listdir(self._PKI_DB_PATH):
-            logger.debug("Initializing db at '%s'", self._PKI_DB_PATH)
+            logger.info("Initializing db at '%s'", self._PKI_DB_PATH)
```
manneohrstrom: I suggest: `Initializing chrome certificate registry '%s'`

python/tk_framework_desktopserver/certificates.py
jfboismenu: The calling method already has extensive logging around what it does. I don't believe this is needed here.

python/tk_framework_desktopserver/certificates.py
jfboismenu: This is what the `ctx` already is.

framework.py
jfboismenu: No, because the dialogs are short lived and not permanent. The `parent` is necessary so the dialogs always appear on top of the Shotgun Desktop, otherwise having the desktop in "On Top" mode would make the popup appear behind the window.

PR #35, opened 2017-06-09
For #43236: Handles empty legacy caches gracefully.

thebeeland:
The end behavior matches wss1, except that logs go to tk-desktop.log instead of the browser console when there is a parsing error.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -1313,28 +1313,32 @@ class ShotgunAPI(object):
 
             # The data returned by the tank command is a newline delimited string
             # that defines rows of ordered data delimited by $ characters.
-            for line in raw_actions_data["out"].split("\n"):
-                action = line.split("$")
+            try:
+                for line in raw_actions_data["out"].split("\n"):
+                    action = line.split("$")
 
-                if action[2] == "":
-                    deny_permissions = []
-                else:
-                    deny_permissions = action[2].split(",")
+                    if action[2] == "":
+                        deny_permissions = []
+                    else:
+                        deny_permissions = action[2].split(",")
 
-                multi_select = action[3] == "True"
+                    multi_select = action[3] == "True"
 
-                commands.append(
-                    dict(
-                        name=action[0],
-                        title=action[1],
-                        deny_permissions=deny_permissions,
-                        supports_multiple_selection=multi_select,
-                        app_name=None, # Not used here.
-                        group=None, # Not used here.
-                        group_default=None, # Not used here.
-                        engine_name=None, # Not used here.
+                    commands.append(
+                        dict(
+                            name=action[0],
+                            title=action[1],
+                            deny_permissions=deny_permissions,
+                            supports_multiple_selection=multi_select,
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1300,51 +1310,56 @@ def _legacy_process_configs(self, config_data, entity_type, project_id, all_acti
...
-            # We don't need or want the descriptor object to be sent to the client.
-            # Since we're done with this config for this invokation, we can just
-            # delete it from the entity dict.
-            del config_entity["descriptor"]
-
-            # And since we know this set of actions came from this legacy path, we
-            # can go ahead and include some extra data in the config dict that we
-            # can key off of when this action is called from the client.
-            config_entity[constants.LEGACY_CONFIG_ROOT] = config_path
+            except IndexError:
```
manneohrstrom: i would prefer a broader catch here - we don't expect ANYTHING to go wrong in here, right? we are not calling any methods and we don't want anything to bubble up. Would also be nice to include what the error is and perhaps not say "legacy".

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1275,6 +1275,16 @@ def _legacy_process_configs(self, config_data, entity_type, project_id, all_acti
             config_path, config_entity = config_data
             commands = []
 
+            # We don't need or want the descriptor object to be sent to the client.
+            # Since we're done with this config for this invokation, we can just
+            # delete it from the entity dict.
+            del config_entity["descriptor"]
```
manneohrstrom: a possibly clearer and more robust way to do this is to build up a dictionary from scratch and include exactly the fields we want to pass to the user rather than subtracting them like this. That approach means we are only making assumptions about the `config_entity` keys that we want to operate on in this method. 

If for example the `descriptor` key was removed or a `foo_bar` key was added to `config_entity`, the logic here would break. 

IME, It is also generally safer to try to be more functional/immutable - avoid deleting things out of dictionaries in favour of building new data structures unless there is a serious performance hit - in that case we can comment why we are doing things that way.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1300,51 +1310,56 @@ def _legacy_process_configs(self, config_data, entity_type, project_id, all_acti
...
+                all_actions[config_name] = dict(
+                    actions=[],
+                    config=config_entity,
+                )
+
                 continue
 
             if raw_actions_data["retcode"] != 0:
                 logger.error(
                     "A shotgun_get_actions call did not succeed: %s",
```
manneohrstrom: What does this error message look like for support? Should we include the yml file path or the config descriptor to make it more helpful?

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: That's fair. I've addressed this in a different branch where I'm currently working.

thebeeland:
Manne's comments have been addressed in a different branch. All of the changes in this PR will also be merged as part of a newer branch once that has been reviewed.

PR #36, opened 2017-06-09
For #43105: Shallow yml mtime statting.

thebeeland:
This change is for speed. Previously we statted all yml files found in the config by traversing the config recursively. Now we go straight to the subset of files that we know we want, and don't worry about anything else.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -62,6 +62,7 @@ class ShotgunAPI(object):
     SOFTWARE_ENTITIES = "software_entities"
     ENTITY_TYPE_WHITELIST = "entity_type_whitelist"
     LEGACY_PROJECT_ACTIONS = "legacy_project_actions"
+    YML_FILE_DATA = "yml_file_data"
 
     def __init__(self, host, process_manager, wss_key):
         """
@@ -117,6 +118,7 @@ class ShotgunAPI(object):
     ###########################################################################
     # Public methods
 
+    @sgtk.LogManager.log_timing
     def execute_action(self, data):
         """
         Executes the engine command associated with the action triggered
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -835,22 +836,30 @@ def _get_contents_hash(self, config_descriptor, entities):
         # the modtimes of all yml files found. When they're added to the digest,
         # we'll dump as json sorting alphanumeric on yml file name to ensure
         # consistent ordering of data.
+        scanned = 0
```
manneohrstrom: we have a timer decorator - i suggest adding this to this method. And do a pass and see if we should add it somewhere else.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -835,22 +836,30 @@ def _get_contents_hash(self, config_descriptor, entities):
...
+            # that this means we are not covering included yml files, but we
+            # accept that for the sake of speed.
+            config_path = os.path.join(config_descriptor.get_path(), "env")
 
             if config_path is not None:
-                for root, dir_names, file_names in os.walk(config_path):
-                    for file_name in fnmatch.filter(file_names, "*.yml"):
-                        full_path = os.path.join(root, file_name)
-                        yml_files[full_path] = os.path.getmtime(full_path)
+                logger.debug("Statting yml files in %s", config_path)
```
manneohrstrom: i think we can improve this log message to make debugging and tech support easier. It tells us what we are doing, but not why. How about something like:

```
DEBUG: Descriptor xxx is mutable - will include the modification time of environment files in the cache 
hash key
```

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -835,22 +836,30 @@ def _get_contents_hash(self, config_descriptor, entities):
...
+
+                for file_path in glob.glob(os.path.join(config_path, "*.yml")):
+                    yml_files[file_path] = os.path.getmtime(file_path)
+                    scanned += 1
```
manneohrstrom: can skip `scanned` and just use `len(yml_files)`.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -835,22 +836,30 @@ def _get_contents_hash(self, config_descriptor, entities):
...
 
             hashable_data["modtimes"] = yml_files
 
+        logger.debug("Files statted for mtime: %s", scanned)
```
manneohrstrom: Presumably `json.dumps` is expensive - why do we need to turn things into a json string before we turn it into a hash number? Similarly, do we need to store every single `file_path` in a dictionary along with its mod time just to eventually arrive at a number. Feels like we are building up a massive big data structure, then parsing it and stringifying it into json only to then walk over that entire string to compute a hash. 

It probably doesn't matter in this case given the number of files, but may still be worth taking a quick look and potentially revisiting this data structure - feels like there is a simpler solution here that scales in a better way.

manneohrstrom: Maybe worth pointing out the shallow-ness in the debug message to help support?
e.g something like

```
DEBUG: computed hash based on 12 config environment yml files. (Shallow scan, not including subfolders.)
```

thebeeland: Agreed re: json.dumps. I'll remove that and just hash the data structure directly.

As for the contents of the data structure, I have newer code that's structured a bit differently than this now. I have a generic lookup of environment file data (file paths, modtimes) which is cached per-wss-connection. The relevant data is extracted from that to compute the contents hash, as well as to provide the list of env files needed to determine the full whitelist of entity types that we support, as we're now adding in those that have a defined shotgun_xxx.yml file.

thebeeland: Correction: now I know why I json encoded the hashable data; dict objects aren't hashable by way of the builtin hash() function or via md5. Given that the hash contains both the file mtime data and the Software entities found in SG, the easiest way to do this is to json encode the data sorted by key, which ensures a matching hash for matching data.

thebeeland: I've commented explaining the why of it.

thebeeland:
Merged as part of a separate PR.

PR #37, opened 2017-06-12
For #43058, 43060: Entity type whitelist now contains shotgun_xxx.yml types.

thebeeland:
* Entity types that have an associated shotgun_xxx.yml file in a mutable config will be added to the entity type whitelist.
* Centralizes, and caches in memory per connection, a yml file data structure. This includes the information necessary to both compute a contents hash for a mutable config, and get the shotgun_xxx.yml file names needed for the first bullet point in this list.
* A handful of small changes that came out of code review for other PRs.
* Fixes the v2 code path's cache keys for Task entities. Previously, there was a bug that caused Task entities to receive actions for their parent entity's type (ie: Shot or Asset). Now, actions are looked up specific to the given Task entity, but the cache key is constructed to include the parent entity's type. This allows for future workflows where different actions can be configured for Tasks linked to different parent entity types. This isn't provided out of the box, given how tk-config-default's and tk-config-basic's pick_environment hooks work, but it's possible to implement with relative ease.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -62,6 +62,7 @@ class ShotgunAPI(object):
     SOFTWARE_ENTITIES = "software_entities"
     ENTITY_TYPE_WHITELIST = "entity_type_whitelist"
     LEGACY_PROJECT_ACTIONS = "legacy_project_actions"
+    YML_FILE_DATA = "yml_file_data"
 
     def __init__(self, host, process_manager, wss_key):
         """
@@ -117,6 +118,7 @@ class ShotgunAPI(object):
     ###########################################################################
     # Public methods
 
+    @sgtk.LogManager.log_timing
     def execute_action(self, data):
         """
         Executes the engine command associated with the action triggered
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -441,10 +406,43 @@ def get_actions(self, data):
...
-                # reply to the client, we need to filter out the descriptor object.
-                # It's neither useful to the client, nor json encodable.
-                del pipeline_config["descriptor"]
+                # We'll rebuild this pipeline_config dict to only include the keys
+                # that we know we want to pass back to the client. In the event that
+                # the interface to getting these config dicts changes in the future,
+                # it will help us keep from passing back uneeded data or, even worse,
+                # data that can't be serialized, which would cause an exception.
+                pipeline_config = dict(
+                    id=pipeline_config["id"],
```
manneohrstrom: nice!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -971,22 +995,34 @@ def _get_lookup_hash(self, config_uri, project, entity_type, entity_id):
...
+        # but it does when tk-shotgun is configured in a non-shotgun_xxx.yml
+        # environment that the config's pick_environment hook recognizes as a
+        # Task entity's target environment. In that case, it would be possible
+        # to configure different engine commands for tk-shotgun when a Task is
+        # linked to a Shot versus when it's linked to an Asset.
+        if entity_type == "Task":
+            logger.debug("Task entity detected, looking up parent entity...")
+            parent_entity_type = self._get_task_parent_entity_type(entity_id)
+            logger.debug("Task entity's parent entity type: %s", entity_type)
+            cache_key += parent_entity_type
```
manneohrstrom: we pass this to the hook right? Do we need to change anything in the hook (param names, docstrings, examples etc) to help illustrate the new task behaviour?

thebeeland: This is happening post-hook-method, so regardless of what happens there, we get the parent entity type appended to the key.

PR #40, opened 2017-07-20
For #42679, can't jump into more recent core

jfboismenu:
The core now returns the path to the root of installed configuration, so we don't need to remove /config from the path.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -14,11 +14,8 @@ import re
 import cPickle
 import sqlite3
 import json
-import subprocess
 import tempfile
 import contextlib
-import json
-import fnmatch
 import datetime
 import copy
 import base64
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -14,11 +14,8 @@
 import cPickle
 import sqlite3
 import json
-import subprocess
```
jfboismenu: I also cleaned up whitespace issues and unused imports.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -421,7 +419,7 @@ def _get_actions(self, data):
                     # path is routed to <root>/config, while the v1 api is just
                     # wanting the root path where the tank command lives.
                     legacy_config_data[config["name"]] = (
-                        os.path.dirname(config["descriptor"].get_path()),
+                        config["descriptor"].get_path(),
```
jfboismenu: The actual fix is here.

PR #41, opened 2017-07-21
For #43032: Fixes for PYTHONPATH and Python intepreter paths.

thebeeland:
* Ensures usage of the config’s stated Python interpreter.
* Extends sys.path in the cache and command subprocesses to ensure we don’t blow away PYTHONPATH entries.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -221,13 +221,23 @@ class ShotgunAPI(object):
             "execute_command.py"
         )
 
-        # We'll need the Python executable when we shell out. We can't
-        # rely on sys.executable, because that's going to be the Desktop
-        # exe on Windows. We'll pull from the site config's interpreter
-        # cfg file.
-        python_exe = sgtk.get_python_interpreter_for_config(
-            self._engine.sgtk.pipeline_configuration.get_path(),
-        )
+        # We'll need the Python executable when we shell out. We need to make
+        # sure it's what's in the config's interpreter config file. To do that,
+        # we can get the config's descriptor object from the manager and ask
+        # it for the path. By this point, all of the config data has been cached,
+        # because it will have been looked up as part of the get_actions method
+        # when the menu asked for actions, so getting the manager and all of the
+        # config data will be essentially free.
+        manager = self._get_toolkit_manager()
+        with self._LOCK:
+            manager.bundle_cache_fallback_paths = self._engine.sgtk.bundle_cache_fallback_paths
+            all_pc_data = self._get_pipeline_configuration_data(
+                manager,
+                project_entity,
+                data,
+            )
+
+        python_exe = all_pc_data[config_entity["id"]]["descriptor"].python_interpreter
         logger.debug("Python executable: %s", python_exe)
 
         args = [python_exe, script, args_file]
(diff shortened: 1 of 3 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -717,17 +720,13 @@ def _cache_actions(self, data, config_data):
...
-        # We'll need the Python executable when we shell out. We can't
-        # rely on sys.executable, because that's going to be the Desktop
-        # exe on Windows. We'll pull from the site config's interpreter
-        # cfg file.
-        python_exe = sgtk.get_python_interpreter_for_config(
-            self._engine.sgtk.pipeline_configuration.get_path(),
-        )
+        # We'll need the Python executable when we shell out. We want to make sure
+        # we use the Python defined in the config's interpreter config file.
+        python_exe = descriptor.python_interpreter
```
jfboismenu: Sweet!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -221,13 +221,23 @@ def execute_action(self, data):
...
-            self._engine.sgtk.pipeline_configuration.get_path(),
-        )
+        # We'll need the Python executable when we shell out. We need to make
+        # sure it's what's in the config's interpreter config file. To do that,
+        # we can get the config's descriptor object from the manager and ask
+        # it for the path. By this point, all of the config data has been cached,
+        # because it will have been looked up as part of the get_actions method
+        # when the menu asked for actions, so getting the manager and all of the
+        # config data will be essentially free.
+        manager = self._get_toolkit_manager()
```
jfboismenu: I notice the lock in other PRs but didn't give much thought to it. Have you considered creating one on the fly instead so each thread has its own? This would remove the need for the lock, right?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -255,7 +255,7 @@ def cache(
     with open(arg_data_file, "rb") as fh:
         arg_data = cPickle.load(fh)
 
-    sys.path = arg_data["sys_path"]
+    sys.path.extend(arg_data["sys_path"])
```
jfboismenu: should we prepend everything instead, to make it takes precedence over anything that is in the path?

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: No, that's not true. The situation isn't that a single manager isn't thread safe, it's that the bootstrapping process in general is not thread safe. Manne can explain better than I can, but given that we're doing (potentially) quite a few filesystem operations to create/manage the bundle cache, concurrent bootstraps can step all over each other.

Manne and I had wanted to be able to run concurrent bootstraps, but all kinds of problems cropped up as a result. The compromise was to allow concurrency where we can, but to lock the operations that were causing issues.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: Good question. I've been going back-and-forth on that in my head since yesterday.

I think I've convinced myself that you're correct, and that it should be prepended. That will ensure that the sgtk imported here prior to bootstrap is the same tk-core as is running in Desktop. I think that's the most important thing.

manneohrstrom: +1 for prepend. I think that's how we have approached in in other places. If toolkit requires something, we should make sure that we get exactly the version of which we expect. Say for example someone has an older version of the python API already in the pythonpath - extending the path would cause errors in toolkit (hard to track down) as it is expecting a more recent version of the API.

python/tk_framework_desktopserver/shotgun/api_v2.py
manneohrstrom: @thebeeland can you write up a ticket where we look at doing the manager reentrant? it seems we should ideally be able to run two tk managers in two different threads without them conflicting.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1358,6 +1357,20 @@ def _get_task_parent_entity_type(self, task_id):
 
         return cache[self.TASK_PARENT_TYPES][task_id]
 
+    def _get_toolkit_manager(self):
+        """
+        Gets an initialized ToolkitManager object.
+
+        :returns: A ToolkitManager object.
+        """
+        if self.TOOLKIT_MANAGER is None:
```
manneohrstrom: Why are we keeping this around? performance? I don't think we gain much in terms of performance so wondering if it's better to instantiate a new object every time?

PR #42, opened 2017-07-28
For #41325, shared secret

jfboismenu:
Complete implementation of the new self-signed certificate-less browser integration.

On startup, we detect if the site supports the browser integration without self-signed certs and use shotgunlocalhost.com certs pulled from Shotgun if that is the case.

Then the protocol checks that we do not allow any queries to any commands while in unencrypted mode.

I've also added tests. Each test resolves around a concept of steps. Each step is an asynchronous bit of code that returns a deferred which Twisted can use to wait before launching the next asynchronous bit. We simply run those steps one after the other. Note that this is different than what is called Deferred chaining, which allows multiple deferreds to be run synchronously, each updating the previous one's results.


python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -77,13 +87,10 @@ def connectionLost(self, reason):
 
             if certificate_error:
                 logger.info("Certificate error!")
-                StatusServerProtocol.serverStatus = StatusServerProtocol.SSL_CERTIFICATE_INVALID
```
jfboismenu: This code is dead. We removed the status server protocol before the initial release of the websocket server.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -140,6 +155,10 @@ def onMessage(self, payload, is_binary):
 
         self._protocol_version = message["protocol_version"]
 
+        if message["command"]["name"] == "activate_encryption":
```
jfboismenu: An attacker could capture the activate_encrption message and never let it be forwarded to our server, which means the server would keep serving data as unencrypted, which needs to be supported for backwards compatibility with previous versions of Shotgun.

To close this hole, we need to make encryption mandatory for the server to handle any other requests than get_protocol_version.

.gitignore
```diff
@@ -7,7 +7,4 @@ __pycache__/
 
 # Twisted Unit Test Artifacts
 _trial_temp*/
-_trial_temp*
-
-# Don't commit keys
-resources/keys/
```
jfboismenu: I removed the standalone browser integration, which also means the resources/keys is not used anymore.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -52,9 +85,6 @@ def protocol_version(self):
         """
         return self._protocol_version
 
-    def onClose(self, was_clean, code, reason):
```
jfboismenu: Removed dead code.

python/tk_framework_desktopserver/server.py
```diff
@@ -106,15 +121,18 @@ def _start_server(self):
...
         self.factory = WebSocketServerFactory(
-            "wss://localhost:%d" % self._port, debug=self._debug, debugCodePaths=self._debug
+            "wss://localhost:%d" % self._port # , debug=self._debug, debugCodePaths=self._debug
         )
 
         self.factory.protocol = ServerProtocol
         self.factory.host = self._host
         self.factory.user_id = self._user_id
         self.factory.notifier = self.notifier
-        self.factory.setProtocolOptions(allowHixie76=True, echoCloseCodeReason=True)
```
jfboismenu: hixie76 adds support for the draft implementation of websocket. that support has been removed from Twisted it seems.

tests/python/base_test.py
```diff
@@ -41,6 +43,12 @@ def setUp(self):
         # now make a context
         context = self.tk.context_from_entity(self.project["type"], self.project["id"])
 
+        patched = patch(
+            "sgtk.platform.Engine._define_qt_base", return_value={"qt_core": Mock(), "qt_gui": Mock()}
```
jfboismenu: Mocks Qt so unit tests can run without Qt installed. :)

tests/server/test_server.py
```diff
@@ -0,0 +1,536 @@
...
+        return self._chain_calls(self._activate_encryption_if_required, step1, step2)
+
+    return type(class_name, class_parents, class_attr)
+
+
+class TestEncryptedServer(unittest.TestCase):
+    """
+    Tests for various caching-related methods for api_v2.
+    """
+    __metaclass__ = TestServerBase
```
jfboismenu: Seems like twisted.trial has issue with unit test base classes, so we're using a metaclass instead to add basic functionality to the test classes.

tests/server/test_server.py
```diff
@@ -0,0 +1,536 @@
...
+    PUBLIC_API_METHODS = ["repeat_value", "missing_attribute"]
+
+    def __init__(self, host, process_manager, wss_key):
+        self._host = host
+
+    def repeat_value(self, payload):
+        self._host.reply({"value": payload["value"] * 3})
+
+
+def TestServerBase(class_name, class_parents, class_attr):
```
jfboismenu: Seems like twisted.trial has issue with unit test base classes, so we're using a metaclass instead to add basic functionality to the test classes.

jfboismenu:
Currently all the tests are failing on Linux because we're missing third parties.

jfboismenu:
Linux tests fail on Travis-ci because our libraries are compiled with 2 bytes per char and travis uses 4 bytes. However, Windows works!

python/tk_framework_desktopserver/process_manager.py
```diff
@@ -13,7 +13,7 @@
 import glob
 from .command import Command
 
-from PySide import QtGui, QtCore
+from sgtk.platform.qt import QtGui, QtCore
```
thebeeland: +1

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -129,56 +181,156 @@ def onMessage(self, payload, is_binary):
...
-            logger.debug("Origin site: %s", origin_network)
-            logger.debug("Origin user: %s", user_id)
+        if not self._validate_user(host_network, user_id, origin_network):
             self.factory.notifier.different_user_requested.emit(self._origin, user_id)
-            self.sendClose(
-                3001,
-                u"You are not authorized to make browser integration requests. "
-                u"Please re-authenticate in your desktop application."
-            )
+            self.sendClose(*self.UNAUTHORIZED_USER)
```
thebeeland: I was going to ask about the structure of all of these pre-baked error tuples, but than I saw this and it makes good sense. Nice!

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -129,56 +181,156 @@ def onMessage(self, payload, is_binary):
...
+        # a file. This will ensure the server is as responsive as possible. Twisted will take care
+        # of the thread.
         reactor.callInThread(
             self._process_message,
             message_host,
             message,
             message["protocol_version"],
         )
 
+    def _validate_user(self, host_network, user_id, origin_network):
```
thebeeland: Missing param docs.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -129,56 +181,156 @@ def onMessage(self, payload, is_binary):
...
+        """
+        Validates if the user from the browser can connect to this server.
+
+        :returns: True if the user can connect, False otherwise.
+        """
+        # If the hosts are different or the user ids are different, report an error.
+        if host_network != origin_network or (user_id is not None and user_id != self.factory.user_id):
```
thebeeland: [minor] My feeble, 10pm brain has trouble parsing this condition because of all of the nots. Any chance you could flip it around and set the if condition up to return True and return False if it fails the validation checks as the else statement?

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -129,56 +181,156 @@ def onMessage(self, payload, is_binary):
...
+            logger.debug("Origin user: %s", user_id)
+            return False
+        else:
+            return True
+
+    def _handle_get_protocol_version(self):
+        """
+        Handles get_protocol_version
+
+        Note: This is mock by the browser integration crash tool, do not change or you will break
```
thebeeland: [minor] Should probably use a `.. note::` here for proper formatting in Sphinx.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -129,56 +181,156 @@ def onMessage(self, payload, is_binary):
...
+
+    def _retrieve_server_secret(self):
+        """
+        Retrieves the server secret from Shotgun.
+
+        .. note::
+            This method caches the result of the retrieval for any other ServerProtocol instances
+            created after the first one.
+
+            Because
```
thebeeland: Good docs, but worried about formatting when it's converted to Sphinx. For each bullet point I would do a a proper sentence with punctuation. Also would check what the "Because" will look like without any other treatment.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -129,56 +181,156 @@ def onMessage(self, payload, is_binary):
...
+                - each server instance has a different server id
+                - retrieving a secret with a different id generates a new secret
+                - we can launch two servers at the same time (the second one won't be able to listen
+                  and will fail however.)
+            , it wouldn't be a good idea to retrieve the server id before successfully listening
```
thebeeland: The leading comma is throwing me off. Does it look correct in Sphinx?

python/tk_framework_desktopserver/sgtk_file_dialog.py
```diff
@@ -10,10 +10,7 @@
 
 import sys
 
-try:
-    from PySide import QtCore, QtGui
-except ImportError:
-    from PyQt4 import QtCore, QtGui
+from sgtk.platform.qt import QtCore, QtGui
```
thebeeland: 10 points for Gryffindor!

thebeeland:
Note that I did not look closely at the tests. I certainly can if you want eyes on it, but you can let me know if that's the case.

python/tk_framework_desktopserver/server_protocol.py
jfboismenu: Good point. Done.

python/tk_framework_desktopserver/server_protocol.py
jfboismenu: Good point. I discovered [this tool](http://rst.ninjs.org/#ICAgICAgICAuLiBub3RlOjoKICAgICAgICAgICAgVGhpcyBtZXRob2QgY2FjaGVzIHRoZSByZXN1bHQgb2YgdGhlIHJldHJpZXZhbCBmb3IgYW55IG90aGVyIFNlcnZlclByb3RvY29sIGluc3RhbmNlcwogICAgICAgICAgICBjcmVhdGVkIGFmdGVyIHRoZSBmaXJzdCBvbmUuCgogICAgICAgICAgICBDb25zaWRlciB0aGUgZm9sbG93aW5nCgogICAgICAgICAgICAtIEVhY2ggc2VydmVyIGluc3RhbmNlIGhhcyBhIGRpZmZlcmVudCBzZXJ2ZXIgaWQKICAgICAgICAgICAgLSBSZXRyaWV2aW5nIGEgc2VjcmV0IHdpdGggYSBkaWZmZXJlbnQgaWQgZ2VuZXJhdGVzIGEgbmV3IHNlY3JldAogICAgICAgICAgICAtIFdlIGNhbiBsYXVuY2ggdHdvIHNlcnZlcnMgYXQgdGhlIHNhbWUgdGltZS4KCiAgICAgICAgICAgIEl0IHdvdWxkbid0IGJlIGEgZ29vZCBpZGVhIHRvIHJldHJpZXZlIHRoZSBzZXJ2ZXIgaWQgYmVmb3JlIHN1Y2Nlc3NmdWxseSBsaXN0ZW5pbmcgb24KICAgICAgICAgICAgdGhlIHBvcnQgc2luY2UgbGF1bmNoaW5nIHRoZSBzZXJ2ZXIgYSBzZWNvbmQgdGltZSBieSBtaXN0YWtlIHdvdWxkIHJlc2V0IHRoZSBzZWNyZXQKICAgICAgICAgICAgcmV0cmlldmVkIGJ5IHRoZSBmaXJzdCBpbnN0YW5jZS4=), pretty cool

python/tk_framework_desktopserver/sgtk_file_dialog.py
jfboismenu: I knew what you were quoting, but I was wondering if there was an [extra meaning](http://www.urbandictionary.com/define.php?term=10%20points%20for%20Gryffindor) in american culture I wasn't aware of.

Gotta love Urban Dictionary.

PR #43, opened 2017-08-29
For #44020: Fix for not receiving a project entity from the My Tasks page.

thebeeland:
Also sorts out some exception handling issues in execute_command, and updates tests to handle the new logic for extracting entities from the web app payload.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -63,6 +63,7 @@ class ShotgunAPI(object):
     ENTITY_TYPE_WHITELIST = "entity_type_whitelist"
     LEGACY_PROJECT_ACTIONS = "legacy_project_actions"
     YML_FILE_DATA = "yml_file_data"
+    ENTITY_PARENT_PROJECTS = "entity_parent_projects"
 
     # We need to protect against concurrent bootstraps happening.
     # This is a reentrant lock because get_actions is recursive
@@ -126,6 +127,32 @@ class ShotgunAPI(object):
 
     @sgtk.LogManager.log_timing
     def execute_action(self, data):
+        """
+        Executes the engine command associated with the action triggered
+        in the client.
+
+        :param dict data: The payload from the client.
+        """
+        # When the v2 protocol is being used, the Shotgun web app makes use of
+        # Javascript Promises to handle the reply from the server, whether there
+        # is success or failure. When getting actions, we want to make sure that
+        # if there was some kind of unhandled exception that there is a proper
+        # reply to the client so that the Promise can be kept or broken, as is
+        # appropriate.
+        try:
+            self._execute_action(data)
+        except Exception:
+            import traceback
+            self.host.reply(
+                dict(
+                    err="Failed to execute action: %s" % traceback.format_exc(),
+                    retcode=constants.COMMAND_FAILED,
+                    out="",
+                ),
+            )
+            logger.exception(format_exc())
+
+    def _execute_action(self, data):
         """
         Executes the engine command associated with the action triggered
         in the client.
@@ -325,6 +352,7 @@ class ShotgunAPI(object):
                     out="",
                 ),
             )
+            logger.exception(format_exc())
 
     def _get_actions(self, data):
         """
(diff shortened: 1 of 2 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -132,6 +133,32 @@ def execute_action(self, data):
 
         :param dict data: The payload from the client.
         """
+        # When the v2 protocol is being used, the Shotgun web app makes use of
+        # Javascript Promises to handle the reply from the server, whether there
+        # is success or failure. When getting actions, we want to make sure that
+        # if there was some kind of unhandled exception that there is a proper
+        # reply to the client so that the Promise can be kept or broken, as is
+        # appropriate.
+        try:
```
jfboismenu: +100000

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -132,6 +133,32 @@ def execute_action(self, data):
...
+            self._execute_action(data)
+        except Exception:
+            import traceback
+            self.host.reply(
+                dict(
+                    err="Failed to execute action: %s" % traceback.format_exc(),
```
jfboismenu: I'd like to get our security experts opinion at some point about returning Python error messages back to the browser. I wonder if this isn't some sort of security issue because you are telling an attacker how he broke your server.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -944,41 +972,119 @@ def _get_entities_from_payload(self, data):
...
+                    entities.append(entity)
+
+            # If we were not given a usable project entity, we can pull it from an
+            # entity we've just extracted. This doesn't cover the case of receiving
+            # multiple entities from the web app from different projects, but this isn't
+            # the only place where we're going to suffer there.
+            if project_entity is None:
+                project_entity = entities[0]["project"]
+                if project_entity is None:
+                    raise RuntimeError("Unable to determine a project entity from data: %s" % data)
```
jfboismenu: Is this likely to happen? If so, could you give us an example why it might happen?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -944,41 +972,119 @@ def _get_entities_from_payload(self, data):
             that order.
```
jfboismenu: It's not clear to me based on the documentation and the code what the payload can be. There seems to be a lot of edge cases in the method and I think it would be nice to provide an example of each type on payloads you can get from the browser.

jfboismenu: I'm getting a feeling that each entity has the field `id` and `type` and sometimes `project`, because the web app has been inconsistent about this in the past maybe? Is that the right interpretation?

jfboismenu: In any case, clearing that up with a comment would be helpful.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -944,41 +972,119 @@ def _get_entities_from_payload(self, data):
...
+        :returns: A standard Shotgun Project entity.
+        :rtype: dict
+        """
+        if entity.get("project") is not None:
+            return entity["project"]
+
+        if entity["type"] == "Project":
+            return entity
+
+        if self.ENTITY_PARENT_PROJECTS not in self._cache:
```
jfboismenu: Have you added this cache because you perceived a performance problem or because you thought it could turn into one someday? An entity move from one project to another would not be detected when getting the parent project because of this cache. Not a big concern, I know, but if performance is not an issue at the moment, maybe its safer to not add a cache for now, while also add a comment about potentially adding one in the future and mention the dangers of doing so.

jfboismenu: With that being said, the pattern of "if missing key/value then insert" can be written as a one liner as

```
project_cache = self._cache.setdefault(self.ENTITY_PARENT_PROJECTS, {})
if entity["id"]  not in project_cache:
```

jfboismenu: If also saves the code from having to reevaluate `self._cache[self.ENTITY_PARENT_PROJECTS]`.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Good point. I'll bring it up with Rob next time he and I have our 1:1.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Will do, and yeah, it's a bit squirrelly and inconsistent. What I was fixing in this PR is yet another inconsistency that comes from one specific section of the My Tasks page.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: Should be unlikely verging on impossible. My only thought here was catching some case of somehow getting a site-level entity that doesn't have a project association, and dying here in some way that gives a decent error message. It's just being (overly) cautious, though.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: It's worth noting that most of the inconsistencies I would consider bugs in the toolkit menu code in the web app. Especially this one here from My Tasks. However, it's significantly easier to work around it in Python than it is to hunt it down in javascript, and it's safer to release, as well.

PR #45, opened 2017-09-06
Reverted sys.path.insert to append  to allow injection of custom binary libs

mbrainerd:
v1.1.x used sys.path.append. v1.2.x uses sys.path.insert. This broke us at DD as we needed to inject a custom libffi.so into the path that was compatible with our runtime environment that the insert stomps on.


mbrainerd:
Any update on this request?

mbrainerd:
Any update on this request?

PR #46, opened 2017-09-28
For #42770: Only replies with a full tb when debug logging is on.

thebeeland:
* When an unhandled exception is raised, we now only format the traceback and send it to the caller if debug logging is on. When debug logging is off, a generic error message is used instead.
* Uses `cgi.escape` to sanitize replies when the string in question isn't fully known.

There will be additional fixes required in the web app, but this should cover the holes in api_v2.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -22,6 +22,8 @@ import base64
 import glob
 import threading
 import hashlib
+import cgi
+import traceback
 
 import sgtk
 from sgtk import TankFileDoesNotExistError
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -161,27 +163,16 @@ def _execute_action(self, data):
...
-                logger.debug("Clone configuration successful.")
-                self.host.reply(
-                    dict(
-                        retcode=constants.COMMAND_SUCCEEDED,
-                        err="",
-                        out="",
-                    ),
-                )
-                return
+            self._clone_configuration(data)
```
thebeeland: Note: This is just cleanup. The handling of the exception here isn't needed any longer now that exception handling is more robust in this method's caller.

PR #47, opened 2017-10-27
For #45375: Implements asynchronous caching.

thebeeland:
Additional changes:
* YML file mtime stats during cache validation is now much deeper than before. We crawl to the bottom of the config's env area.
* A sys.path bug in the caching and execution subprocesses is now resolved. We are now guaranteed to get the correct tk-core imported when the subprocesses are spawned.

The intent of these changes are to address issues raised during client testing (the sys.path issue above), and those found during QA of tk-config-default2 (the yml file mtime depth). The changes to how deep we traverse the config looking for yml files to stat necessitates running the caching validation and potential recache asynchronously. The reason for this is that on slow network storage, the time needed to crawl the config and stat the relevant yml files can be quite slow.

The solution here resolves any speed issues that might arise, but the approach opens us up to a known regression in UX. Because the cache validity is checked asynchronously, the first time a user shows the toolkit menu after a config change, they will receive old data. Subsequent requests for menu actions will then be correct, as the cache will have been updated asynchronously the first time the menu was shown. This is regression that we're willing to accept given the speed benefits of the approach.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -24,6 +24,8 @@ import threading
 import hashlib
 import cgi
 import traceback
+import time
+import fnmatch
 
 import sgtk
 from sgtk import TankFileDoesNotExistError
@@ -50,6 +52,11 @@ class ShotgunAPI(object):
         "pick_files_or_directories",
     ]
 
+    # Stores cache keys that have been checked validated and at
+    # what time that occurred.
+    CACHE_VALIDATED = dict()
+    CACHE_VALIDATION_INTERVAL = 2.0 # Seconds
+
     # Stores data persistently per wss connection.
     WSS_KEY_CACHE = dict()
     DATABASE_FORMAT_VERSION = 1
@@ -66,6 +73,7 @@ class ShotgunAPI(object):
     LEGACY_PROJECT_ACTIONS = "legacy_project_actions"
     YML_FILE_DATA = "yml_file_data"
     ENTITY_PARENT_PROJECTS = "entity_parent_projects"
+    SHOTGUN_YML_FILES = "shotgun_yml_files"
 
     # We need to protect against concurrent bootstraps happening.
     # This is a reentrant lock because get_actions is recursive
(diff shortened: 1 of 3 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -545,92 +553,83 @@ def _get_actions(self, data):
...
-                        # In that case, we just continue on with cached_data set
-                        # to an empty list, which will cause the caching subprocess
-                        # to be spawned, which will setup the database and populate
-                        # with the data we need. The behavior as a result of this
-                        # is the same as if we ended up with a cache miss or an
-                        # invalidated result due to a contents_hash mismatch.
-                        logger.debug(
-                            "Cache query failed due to missing table. "
-                            "Triggering caching subprocess..."
+                try:
```
thebeeland: NOTE: This block of change is mostly just due to the tabbing changing after removing the lock context.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -711,23 +710,65 @@ def _compute_sys_path(self):
...
+            work. This represents the situation where we've been asked to
+            re-cache actions, but we then prove that the existing cached data
+            is still valid.
+        :param bool async: Indicates whether the caching routine should be
+            run asynchronously.
         """
+        descriptor = config_data["descriptor"]
+        lookup_hash = config_data["lookup_hash"]
+
+        if async:
```
thebeeland: This is where the thread is spawned that will check the validity of the cached data and then potentially recaching.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1524,6 +1586,42 @@ def _get_toolkit_manager(self):
         return self.TOOLKIT_MANAGER
 
     @sgtk.LogManager.log_timing
+    def _get_shotgun_yml_files(self, config_descriptor):
```
thebeeland: A shallow dive into the config is still needed when building the list of entity types we want to provide actions for. That relies on the existence of any shotgun_xxx.yml environments that might exist, so a separate method is now provided to get only that list of files.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1549,55 +1647,64 @@ def _get_yml_file_data(self, config_descriptor):
...
 
             logger.debug("Files checked for mtime: %s", sorted(yml_files.keys()))
-
             self._cache.setdefault(self.YML_FILE_DATA, dict())[root_path] = yml_files
         else:
             logger.debug("Cached yml file data found for %r.", config_descriptor)
 
         return self._cache[self.YML_FILE_DATA].get(root_path)
 
+    def _get_config_env_root(self, config_root_path):
```
thebeeland: This is just making some common logic shared between the two yml file getters above reusable.

manneohrstrom:
looking now!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -720,14 +719,56 @@ def _compute_sys_path(self):
...
     @sgtk.LogManager.log_timing
-    def _cache_actions(self, data, config_data):
+    def _cache_actions(self, data, config_data, cached_contents_hash=None, async=False):
         """
         Triggers the caching or recaching of engine commands.
 
         :param dict data: The data passed down from the wss client.
         :param dict config_data: A dictionary that contains, at a minimum,
             "lookup_hash", "contents_hash", "descriptor", and "entity" keys.
+        :param str cached_contents_hash: If given, represents the currently
```
manneohrstrom: we should put a big comment in the docstring, saying that this method runs in a thread!

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -720,14 +719,56 @@ def _compute_sys_path(self):
     # Internal methods
 
     @sgtk.LogManager.log_timing
-    def _cache_actions(self, data, config_data):
+    def _cache_actions(self, data, config_data, cached_contents_hash=None, async=False):
```
manneohrstrom: are we overcomplicating the design by hacing `_cache_actions` both somehow being synchronous and async - by calling itself in another thread? Threading code is always brittle, because threading is brittle, so i think by making code running in threads as small and clear as possible, we reduce the risk for future errors. Maybe this method could be split into several different ones?

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -257,7 +255,25 @@ def cache(
...
+
+    try:
+        # The RPC api that spawned us has ensured that the first path
+        # in the sys.path list that it provided us is the root path that
+        # will get us the tk-core that we want. We can't prepend any more
+        # than that, because we might be running in a different version
+        # of Python in this process than SG Desktop is. We don't want to
+        # force Python 2.7 standard library modules to be imported if
+        # we're in Python 2.6 here, as an example.
+        sys.path = [arg_data["sys_path"][0]] + sys.path
```
manneohrstrom: for clarity, would it make sense that `arg_data["sys_path"][0]` was sent down as a separate key, e.g. `arg_data["sgtk_path"]` or something like that?

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -720,14 +719,56 @@ def _compute_sys_path(self):
...
+                time_since = now - self.CACHE_VALIDATED[lookup_hash]
+                if time_since < self.CACHE_VALIDATION_INTERVAL:
+                    logger.debug(
+                        "Recaching of data for %s has already been initiated. "
+                        "This thread will exit without triggering a recache of "
+                        "actions for this entry.", lookup_hash
+                    )
+                    return
+                else:
+                    self.CACHE_VALIDATED[lookup_hash] = now
```
thebeeland: I see some redundant logic here with setting the cached time. I'll clean it up before merging.

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
```diff
@@ -257,7 +255,25 @@ def cache(
...
+        import sgtk
+    finally:
+        sys.path = original_sys_path + arg_data["sys_path"]
```
manneohrstrom: how self contained is this script? Why exactly do we even need to append `arg_data["sys_path"]` at all? The comments are great, and indeed very verbose, yet i think it would still be useful with a concrete example of why we need `arg_data["sys_path"]` to be appended to this process' `sys.path`

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -359,7 +353,26 @@ def execute(config, project, name, entities, base_configuration, engine_name, bu
...
+
+    try:
+        # The RPC api that spawned us has ensured that the first path
+        # in the sys.path list that it provided us is the root path that
+        # will get us the tk-core that we want. We can't prepend any more
+        # than that, because we might be running in a different version
+        # of Python in this process than SG Desktop is. We don't want to
+        # force Python 2.7 standard library modules to be imported if
+        # we're in Python 2.6 here, as an example.
+        sys.path = [arg_data["sys_path"][0]] + sys.path
```
manneohrstrom: see comments made in other script.

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: It seems like it can be split relatively easily. Sounds like a good idea to me!

python/tk_framework_desktopserver/shotgun/scripts/cache_commands.py
thebeeland: I think you're almost certainly correct that it's overkill. I'm shifting to only passing over the tk-core root path so that sgtk can be imported, and not doing anything else with sys.path.

PR #48, opened 2017-10-30
update to use api method for core python location

mbrainerd:
Updating to use get_core_python_location() api method instead of assuming path hierarchy.


mbrainerd:
Any update on this request?

jfboismenu:
Hi @mbrainerd , sorry this fell completely off our radar. We're terribly sorry. Did you encounter any bugs? What was the issue exactly?

mbrainerd:
At Digital Domain, due to the specifics of our package management software, our core python installation location is at a different relative path. We have a fork of tk-core that overrides get_core_python_location() but cannot override get_install_location() without touching a much larger swath of core code.

PR #49, opened 2018-01-19
For #46128, multiple endpoints in toolkit.ini

jfboismenu:
Certain locally hosted clients have multiple endpoints for Shotgun. For example, they might set up an endpoint for webapp access and another one for API access, allowing them to spread the query load on multiple servers.

This becomes an issue because a Shotgun user browsing `https://localshotgunserver` but who uses the `https://localapiserver` with the Shotgun Desktop will not be able to answer browser integration requests, since the names won't match.

To accomodate that setup, we're introducing a list of `HostAliases` to toolkit.ini, which allows to define a list of mappings beween a host and other endpoints. Here's an example:

```
[HostAliases]
localhostserver=localapiserver,localapiserver2
```

This would allow a Shotgun Desktop running the site config on `localapiserver` or `localapiserver2` to still be able to answer browser integrationr requests from localhostserver.

> Note that the tests are currently failing because I'm leveraging functionality that hasn't been merged in tk-core's master branch. This is one of those cases where I say "Trust me, it works on my machine." :)

### DIFF
--- a/framework.py
+++ b/framework.py
@@ -135,6 +135,12 @@ class DesktopserverFramework(sgtk.platform.Framework):
         return os.path.join(self.cache_location, "keys")
 
     def _get_host_aliases(self, host):
+        """
+        Returns a list of valid hosts that can connect to the browser integration. The returned
+        list only contains the hostname. The port number and protocol are removed.
+
+        :returns: List of hostnames.
+        """
         self.logger.debug("Looking for an alias for host %s.", host)
         # parse the host and keep only the network location, no need for the rest.
         parsed_host = urlparse.urlparse(host)
@@ -142,7 +148,7 @@ class DesktopserverFramework(sgtk.platform.Framework):
         # members are not None, in which case we want just the hostname and don't
         # care about the port number. If hostname is not set, then we can grab
         # the network location safely.
-        hostname = parsed_host.hostname or parsed_host.netloc.lower()
+        hostname = (parsed_host.hostname or parsed_host.netloc).lower()
         self.logger.debug("Hostname is %s.", hostname)
 
         # Return the dictionary into a list of pool of aliases. Each list is
--- a/framework.py
+++ b/framework.py
@@ -118,6 +118,7 @@ class DesktopserverFramework(sgtk.platform.Framework):
                 encrypt=encrypt,
                 host=host,
                 user_id=user_id,
+                alternative_hosts=self._settings.alternative_hosts,
                 port=self._settings.port
             )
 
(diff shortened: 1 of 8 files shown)


framework.py
```diff
@@ -185,8 +218,21 @@ def __retrieve_certificates_from_shotgun(self):
         self.logger.debug("Retrieving certificates from Shotgun")
         certs = self.shotgun._call_rpc("sg_desktop_certificates", {})
         sgtk.util.filesystem.ensure_folder_exists(self._get_shotgunlocalhost_keys_folder())
-        self._write_cert("server.crt", certs["sg_desktop_cert"])
-        self._write_cert("server.key", certs["sg_desktop_key"])
+        if not certs["sg_desktop_cert"]:
```
jfboismenu: Locally hosted sites by default do not have the certs installed, so we should log that error and ask clients to contact support.

python/tk_framework_desktopserver/settings.py
```diff
@@ -35,122 +34,97 @@ class Settings(object):
     _PORT_SETTING = "port"
     _CERTIFICATE_FOLDER_SETTING = "certificate_folder"
     _ENABLED = "enabled"
+    _HOST_ALIASES = "HostAliases"
 
-    def __init__(self, location, default_certificate_folder):
+    def __init__(self, default_certificate_folder):
```
jfboismenu: Since the browser integration API is not usable in standalone mode anymore, we can remove that bit of backwards compatibility.

tests/api_v2/test_cache.py
```diff
@@ -12,10 +12,6 @@
 import sys
 
 from tank_test.tank_test_base import setUpModule
-
-# import the test base class
-test_python_path = os.path.abspath(os.path.join( os.path.dirname(__file__), "..", "python"))
```
jfboismenu: `run_tests.py` adds the `python` folder of other repos now, no need to do that. :)

tests/server/test_server.py
```diff
@@ -57,31 +57,36 @@ def repeat_value(self, payload):
         self._host.reply({"value": payload["value"] * 3})
 
 
-def TestServerBase(class_name, class_parents, class_attr):
+class TestServerBase(unittest.TestCase):
```
jfboismenu: There's a lot of noise here, but that 's simply refactoring so TestServerBase can be reused for the new unit tests. I recommend scrolling down to the bottom of the file to see the new tests.

jfboismenu:
@robblau I've added you just FYI. No need to actually review. The merge comments should be enough.

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -242,30 +243,33 @@ def _validate_user(self, user_id):
         # origin is formatted such as https://xyz.shotgunstudio.com:port_number
         # host is https://xyz.shotgunstudio.com:port_number
         host_network = self.factory.host.lower()
-        origin_network = self._origin.lower()
+        parsed_host = urlparse.urlparse(self._origin)
+        origin_network = (parsed_host.hostname or parsed_host.netloc).lower()
```
thebeeland: nitpick: probably good to include the same comment here as above where you also did the fallback to netloc.

python/tk_framework_desktopserver/settings.py
thebeeland: 👍

python/tk_framework_desktopserver/settings.py
```diff
@@ -35,122 +34,97 @@ class Settings(object):
...
+        # Also skip empty tokens.
+        for main_host, secondary_hosts in raw_host_aliases.iteritems():
+            main_host = main_host.strip().lower()
+            # Skip empty hosts.
+            if not main_host:
+                continue
+
+            self._host_aliases[main_host] = [
+                secondary_host.lower().strip()
+                for secondary_host in secondary_hosts.split(",")
```
thebeeland: nitpick: might could do this in a single go with re.split splitting on r"[,\s]+"

jfboismenu: I guess we could do that, but we would still require the list comprehension to put anything in lower case. I feel like invoking strip() is clearer to the reader than a regex however.

PR #52, opened 2018-10-03
SG-4611: Better error handling around local file link opening, and more debug logging.

thebeeland:
We're in the process of trying to debug an intermittent local file link open bug that we're having trouble reproducing on our end. The plan right now is to handle the error case more gracefully than before, plus log additional information about which local storages the page making the open request had access to at the time the request was made.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -669,8 +669,14 @@ class ShotgunAPI(object):
             # happen, but in the cases we're attempting to debug it shouldn't be the 
             # case.
             if filepath is None:
-                logger.warning("Shotgun requested a file open via local file linking, "
+                logger.warning(
+                    "Shotgun requested a file open via local file linking, "
                     "but the provided file path is None."
+                )
+            else:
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -655,8 +655,31 @@ def open(self, data):
...
+            filepath = data.get("filepath", "")
+            local_storages = data.get("local_storages")
+
+            # Logging here is for debugging purposes. We have a situation where some
+            # clients are reporting errors when opening files from Shotgun, and the
+            # error implies that we're getting a null value for the file path passed
+            # down from the web app. There are reasonable situations where this might
+            # happen, but in the cases we're attempting to debug it shouldn't be the 
+            # case.
+            if filepath is None:
```
jfboismenu: It might be worth to `logger.debug` the path in the `else:` clause of this `if` statement. This way it puts the next `log.debug` call more in perspective instead of simply knowing it's the "current" one.

thebeeland: 👍

PR #53, opened 2018-10-11
For SG-4623, software with multiple projects are not actionable

jfboismenu:
This fixes a bug where software that is assigned to multiple projects always gets filtered out. This happened because the previous code would filter out an action as soon as the project id wouldn't match one the the projects in the list, effectively preventing any combination of project to match.

I've also update the server tests so that they can be run while the Shotgun Desktop is running, since I kept making that mistake.

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -1632,7 +1632,7 @@ class ShotgunAPI(object):
         # We'll deepcopy the data before returning it. That will ensure that
         # any destructive operations on the contents won't bubble up to the
         # cache.
-        return cache[self.SOFTWARE_ENTITIES]
+        return copy.deepcopy(cache[self.SOFTWARE_ENTITIES])
 
     @sgtk.LogManager.log_timing
     def _get_task_parent_entity_type(self, task_id):
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -60,6 +60,9 @@ class ShotgunAPI(object):
     # Stores data persistently per wss connection.
     WSS_KEY_CACHE = dict()
     DATABASE_FORMAT_VERSION = 1
+    # When the layout of the cache in a cache entry changes, bump this version
+    # so we invalid all cached entries.
+    CACHE_ENTRY_SCHEMA_VERSION = 1
     SOFTWARE_FIELDS = ["id", "code", "updated_at", "type", "engine", "projects"]
     TOOLKIT_MANAGER = None
 
@@ -152,7 +155,7 @@ class ShotgunAPI(object):
         # appropriate.
         try:
             self._execute_action(data)
-        except Exception, e:
+        except Exception:
             self.host.reply(
                 dict(
                     err=self._get_exception_message(),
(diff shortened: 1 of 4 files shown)


tests/api_v2/test_util_methods.py
```diff
@@ -1,23 +1,21 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import os
 import sys
 
-from tank_test.tank_test_base import setUpModule
+from tank_test.tank_test_base import setUpModule # noqa
 
-# import the test base class
```
jfboismenu: tk-core's `run_test.py` automatically adds the unit test's `tests/python` submodule to `sys.path`

jfboismenu:
I had to rewrite the fix in order to support a third bug, which happened when the same engine was configured multiple times for different projects. It used to be that the launchers would be all filtered out. Now they won't, but this behaviour will work only if you pull in [the changes for the launch app](https://github.com/shotgunsoftware/tk-multi-launchapp/pull/63) as well.

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -978,26 +1008,39 @@ def _filter_by_project(self, actions, sw_entities, project):
...
-            # is no filtering to be done, as it's accepted by all projects.
-            for sw in associated_sw:
-                for sw_project in sw.get("projects", []):
-                    if sw_project["id"] != project["id"]:
-                        logger.debug("Action %s filtered out due to SW entity projects.", action)
-                        filtered.append(action)
-                        break
-                if action in filtered:
-                    break
+            if "engine_name" not in action:
```
jfboismenu: I flipped the logic of this loop because we could turn a triple nested loop into a single loop. :)

python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1956,13 +2002,13 @@ def _process_commands(self, commands, project, entities):
         # filter out things like the "Reload and Restart" command.
         filtered = list()
 
-        for command in commands:
-            if command["app_name"] is not None:
-                logger.debug("Keeping command %s -- it has an associated app.", command)
-                filtered.append(command)
+        for cmd in commands:
```
jfboismenu: The `command` variable was shadowing the `command` module imported at the top.

tests/api_v2/test_util_methods.py
```diff
@@ -218,21 +213,30 @@ def test_filter_by_project(self):
         """
         sw = dict(
             code="Tester",
-            engine="tk_engine_tester",
+            engine="tk-maya",
```
jfboismenu: I found tk_engine_tester<number> to be confusing, so I changed those to nice graspable names

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: 👍

python/tk_framework_desktopserver/shotgun/api_v2.py
thebeeland: 👍

tests/server/test_server.py
```diff
@@ -44,6 +46,17 @@
 base.DelayedCall.debug = True
 
 
+# Source: https://stackoverflow.com/a/45690594/10429444
+def find_free_port():
```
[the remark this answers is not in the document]
thebeeland: 👍

PR #61, opened 2020-02-21
SG-14617: Partial Python 3 support for tk-desktop

jfboismenu:
Fixes an import that fails under python 3. This framework is loaded but not actually used when launching a project that might be using Python 3, so we only need to fix that import error.

The rest of the code will be ported when we port the site portion of the tk-desktop engine.

### DIFF
--- a/framework.py
+++ b/framework.py
@@ -12,7 +12,12 @@ import sgtk
 import sys
 import os
 import struct
-import urlparse
+
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse
+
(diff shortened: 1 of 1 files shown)


framework.py
```diff
@@ -12,7 +12,12 @@
 import sys
 import os
 import struct
-import urlparse
+
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse
```
pscadding: Could you use `six.moves.urllib` instead?
https://six.readthedocs.io/#module-six.moves.urllib.parse

PR #64, opened 2020-06-04
Fixes to shotgun integration for python3 projects

eshokrgozar:
Added a test to make sure get_actions and execute_action commands work on both python versions. I need to add the windows portion of the python path discovery and I was thinking of adding a few known paths to search through as well. I figured I'd get the rest CRed while I'm doing that in case there is a better way to accomplish what I'm doing.


tests/python/integration_test.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # Create a user and connection to Shotgun.
+        sa = sgtk.authentication.ShotgunAuthenticator()
+        cls.user = tk_toolchain.authentication._get_toolkit_user(sa, os.environ)
+        cls.sg = cls.user.create_sg_connection()
+
+        # Set the current user in memory for the server and in the file system for the script that
+        # runs the actions in a separate process
+        sgtk.set_authenticated_user(cls.user)
+        sgtk.authentication.session_cache.set_current_host(cls.user.host)
```
jfboismenu, line 81: Ah! I think you're calling this (private) API so that the subprocess knows which user it is right? That's because there's a design flaw in how credentials are passed down to the subprocess. Could you fix that?

The [arguments file](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/master/python/tk_framework_desktopserver/shotgun/api_v2.py#L230) should pass along the credentials for the user and the `cache_commands` and `execute_command` should read the user from that file.

This way you do not have to rely on a global "variable". [Globals.... BAAAAAADDDD!](https://www.youtube.com/watch?time_continue=125&v=fS6udST6lbE&feature=emb_logo)

jfboismenu: You can use use [`serialize_user` and `deserialize_user`](https://developer.shotgunsoftware.com/tk-core/authentication.html#sgtk.authentication.deserialize_user).

tests/python/integration_test.py
```diff
@@ -0,0 +1,206 @@
...
+
+import unittest2
+
+import sgtk
+from sgtk.util import sgre as re
+from sgtk.util.filesystem import safe_delete_folder
+import tk_toolchain.authentication
+
+
+class DesktopServerIntegrationTest(unittest2.TestCase):
```
jfboismenu, line 30: Any reason why we don't use the SgtkIntegrationTest class? When we wrote it it was meant to be reused, but we never had the opportunity until now.

framework.py
```diff
@@ -359,7 +360,7 @@ def __warn_for_prompt(self, parent):
         """
         from sgtk.platform.qt import QtGui
 
-        if sys.platform == "darwin":
+        if platforms.is_macos():
```
jfboismenu, line 363: Use `sgtk.util.is_macos()`, the platforms file is an implementation detail.

python/tk_framework_desktopserver/__init__.py
```diff
@@ -11,16 +11,18 @@
 import os
 import sys
 
+from sgtk.util import platforms
```
jfboismenu, line 14: again

tests/fixtures/config/interpreter_test/env/project.yml
```diff
@@ -0,0 +1,27 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+engines:
+  test_engine_with_frameworks:
+    location:
+      type: path
+      path: ${BUNDLES_ROOT}/test_engine_with_frameworks
```
jfboismenu, line 15: Why can't you use  `$SHOTGUN_TEST_ENGINE`?

eshokrgozar: I didn't notice that we had a test_engine in tk-toolchain. I switched over to using it. I had my frameworks associated with the engine but it was easy enough to move them over to the app.

eshokrgozar:
It looks like the build is using the "run_tests" script which is failing. But the tests pass if you run them with pytest (using tk-toolchain). I'm not sure if I need to update my assumptions to work in both setups or if we just need to update the build script

jfboismenu:
Sorry, I should have made myself clearer the other day. The tests can only be run on Azure if your code fully supports Python 3, which desktopserver doesn't at the moment. So you'll have to put `run_tests: False` for now and leave the old CI in place. Once the framework is fully Python 3 compliant, we can move  the CI entirely to Azure Pipelines.

python/tk_framework_desktopserver/certificates.py
```diff
@@ -20,6 +20,7 @@
 from OpenSSL import crypto
 
 import sgtk
+import sgtk.util
```
jfboismenu, line 23: No need for this second import, the first one [does all the work](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/__init__.py#L117) 😄

PR #66, opened 2020-07-14
SG-18093 Make sure errors are reported

eshokrgozar:
When desktopserver executes actions, exceptions can be hidden from the user because we are filtering out anything that doesn't have our pre-defined log tag. This change catches generic exceptions and logs the error, which means the formatter adds the correct tag, and the filter passes the traceback through to the user (displayed in the webapp). 

I've based the changes off of my python3 branch because the python 3 changes will likely be merged before this and this way the PR will only have the changes relevant to this specific ticket.


eshokrgozar:
Yes. The error will show up in the web app's pop up. It shows up with all the normal, ugly, html escape characters. But there's a separate ticket for that somewhere.

PR #67, opened 2020-07-16
Switched to using sgtk.util.json.load to avoid unicode in python2

eshokrgozar:
We were using the regular json module and getting back unicodes. Passing the base_configuration descriptor as unicode was causing issues down the line because we only check against the descriptor being a string, then convert it to a dictionary. The unicode was falling through the cracks and we ended up with unicode when we were expecting a dict.


jfboismenu:
Do not mind my other review. This is good. These other places have always called `json.loads` from the global Python module and have never been an issue before. If it ain't broken, don't fix it. 😄

PR #68, opened 2020-07-16
SG-16894 Null lookup hash in cache

eshokrgozar:
In my previous commit (which is now merged) I took the single loop that iterates through pipeline configurations and caches the actions and split it into 3 loops. This was done to avoid holding a lock on the cache database the entire time.

Pass 1: Collect the info needed to do a cache lookup for all the configs. 
Pass2: Read the cache for all the keys collected in pass 1
Pass3: Either use the cached value or cache new values for each config

The issue is that when Pass 1 skipped a certain config because it doesn't support the current entity type, in the original code all the rest of the steps would be skipped as well because they were part of the same loop. In the 3 pass setup, the second and third loop were still going through all configs (even the ones that were skipped in the first pass) and we were writing NULL values into the DB and getting stuck in an infinite loop.

This fix makes sure that the other passes also skip the configs that don't support the entity type

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -1397,13 +1397,6 @@ class ShotgunAPI(object):
         :returns: A set of lowercased string entity types.
         :rtype: set
         """
-
-        import pydevd_pycharm
-
-        pydevd_pycharm.settrace(
-            "localhost", port=1234, stdoutToServer=True, stderrToServer=True
-        )
-
         if project_id is None:
             logger.debug("Project id is None, looking up site schema.")
             project_entity = None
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -590,6 +590,11 @@ def _get_actions(self, data):
                 logger.debug("Querying: %s", lookup_hash)
 
                 pc_data["cached_data"] = []
+
+                # If the config doesn't support the current entity_type we don't need to cache it
+                if not lookup_hash:
```
pscadding, line 595: I'm being picky, and this is personal preference so honestly feel free to take it or leave it, but I prefer `if lookup_has is None:` as it is more explicit about what exactly we are checking.

Also about line 589 `lookup_hash = pc_data["lookup_hash"]`.

I know this is the first thing that is defined in the phase 1 loop, but it didn't mean I had to go and check to see if it was always defined. Maybe there is a good reason not to, but I usually go with `pc_data.get("lookup_hash")` and then I wouldn't need to ensure that it is defined.

PR #69, opened 2020-07-31
SG-18619 Fix for incorrect PYTHONPATH when launching Maya

eshokrgozar:
Clients are running into an issue where the PYTHONPATH contains the wrong copy of sgtk in front of the correct one. This means that the wrong hooks will be searched for/loaded when they launch Maya. 

The execute_command file was recently updated to use some core utilities and the imports for these were at the top of the file. To accommodate this, the sgtk path was being added to PYTHONPATH before calling execute_command. 

In this PR I moved sgtk related imports in execute_command to run after we have modified sys.path correctly to include sgtk. At this point there will be no requirement for sgtk to be in the path before calling execute_method so I also removed the addition of sgtk to PYTHONPATH before calling execute_command in api_v2

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
+++ b/python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
@@ -394,7 +394,7 @@ if __name__ == "__main__":
         sys.path = original_sys_path
 
     # Now that we have sgtk.util loaded, use it to make sure we have only utf-8 data in the args
-    arg_data = sgtk.util.json.loads(json.dumps(arg_data))
+    arg_data = sgtk.util.unicode.ensure_contains_str(arg_data)
 
     LOGGING_PREFIX = arg_data["logging_prefix"]
 
--- a/python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
+++ b/python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
@@ -389,8 +389,6 @@ if __name__ == "__main__":
     try:
         sys.path = [arg_data["sys_path"]] + sys.path
         import sgtk
-        import sgtk.util
-        from sgtk.authentication import deserialize_user
         from tank_vendor import six
     finally:
         sys.path = original_sys_path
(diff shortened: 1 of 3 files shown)


python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -382,7 +380,7 @@ def execute(
     arg_data_file = sys.argv[1]
 
     with open(arg_data_file, "rt") as fh:
-        arg_data = sgtk.util.json.load(fh)
+        arg_data = json.load(fh)
 
     # The RPC api has given us the path to its tk-core to prepend
     # to our sys.path prior to importing sgtk. We'll prepent the
```
pscadding, line 386:
```suggestion
    # to our sys.path prior to importing sgtk. We'll prepend the
```

python/tk_framework_desktopserver/shotgun/scripts/get_commands.py
```diff
@@ -235,6 +229,9 @@ def cache(
     finally:
         sys.path = original_sys_path
 
+    # Now that we have sgtk.util loaded, use it to make sure we have only utf-8 data in the args
+    arg_data = sgtk.util.json.loads(json.dumps(arg_data))
```
pscadding, line 233: My feeling here is that it would be better to use `sgtk.util.unicode.ensure_contains_str` on the `arg_data` since we've already had to load it using the standard library, rather than dumping and loading again.

pscadding: https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/util/json.py#L96

eshokrgozar: That makes perfect sense. I made the change

python/tk_framework_desktopserver/shotgun/scripts/execute_command.py
```diff
@@ -382,18 +380,22 @@ def execute(
...
     original_sys_path = copy.copy(sys.path)
     try:
         sys.path = [arg_data["sys_path"]] + sys.path
         import sgtk
+        from tank_vendor import six
     finally:
         sys.path = original_sys_path
 
+    # Now that we have sgtk.util loaded, use it to make sure we have only utf-8 data in the args
+    arg_data = sgtk.util.unicode.ensure_contains_str(arg_data)
```
jfboismenu, line 397: We should have called sgtk.util.json.load(fh) the second time around. `unicode` module is private.

PR #70, opened 2020-08-18
SG-18875: localhost integration fixes under Python 3

jfboismenu:
This fixes a set of issues with the certificate registration process under Python 3. I've added a test to reproduce the problem and then fix it. Unfortunately, it can't run on CI as it requires user intervention from the OS. I'm making sure it always run locally at least so this way we get some assurance that it won't be broken.

### DIFF
--- a/tests/python/base_test.py
+++ b/tests/python/base_test.py
@@ -46,13 +46,6 @@ class TestDesktopServerFramework(TankTestBase):
         # now make a context
         context = self.tk.context_from_entity(self.project["type"], self.project["id"])
 
-        patched = patch(
-            "sgtk.platform.Engine._define_qt_base",
-            return_value={"qt_core": Mock(), "qt_gui": Mock()},
-        )
-        patched.start()
-        self.addCleanup(patched.stop)
-
         # and start the engine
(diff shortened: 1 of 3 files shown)


tests/python/base_test.py
```diff
@@ -46,13 +46,6 @@ def setUp(self):
         # now make a context
         context = self.tk.context_from_entity(self.project["type"], self.project["id"])
 
-        patched = patch(
```
jfboismenu, line 49: I've removed this mock. Let's assume Qt is always available, like on our CI instances, and then we can use Qt in our tests.

tests/test_certificates.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Ensure certificates registration and creation work.
+        """
+
+        from sgtk.platform.qt import QtGui
+
+        if QtGui.QApplication.instance() is None:
+            self.app = QtGui.QApplication([])
+
+        QtGui.QMessageBox.warning(None, "", "You will be prompted to accept updates to the keychain during this test.")
```
jfboismenu, line 34: Display a message to the user to warn them that they will be prompted for their credentials/fingersprint/click on ok by the OS, to avoid developers getting confused by pop-ups during the tests.

PR #74, opened 2020-11-09
SG-19789 Updated certifi version

eshokrgozar:
Updated certifi version to match the one released with Desktop. Added requirements file for the tests directory.


jfboismenu (approved):
I think we need to update the azure-pipelines file to use the new naming convention for credentials.

PR #77, opened 2020-11-26
Add Playlist entity type to the base set of supported types.

staceyoue:
I've renamed the variable `whitelist` to`'allow list` -- do we want to do this with all instances of whitelist?

### DIFF
--- a/python/tk_framework_desktopserver/shotgun/api_v2.py
+++ b/python/tk_framework_desktopserver/shotgun/api_v2.py
@@ -1400,7 +1400,7 @@ class ShotgunAPI(object):
             # environment. Those files contain entity type names that are lower cased,
             # so it's easiest just to do everything that way.
             type_whitelist = set(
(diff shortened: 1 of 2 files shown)


python/tk_framework_desktopserver/shotgun/api_v2.py
```diff
@@ -1400,7 +1400,7 @@ def _get_entity_type_whitelist(self, project_id, config_descriptor):
             # environment. Those files contain entity type names that are lower cased,
             # so it's easiest just to do everything that way.
             type_whitelist = set(
-                [t.lower() for t in constants.BASE_ENTITY_TYPE_WHITELIST]
+                [t.lower() for t in constants.BASE_ENTITY_TYPE_ALLOW_LIST]
```
jfboismenu, line 1403: Thanks for taking the time to update the variable name. We really appreciate it.

thebeeland (approved):

PR #79, opened 2020-12-11
Gracefully dealing with pick_environment returning None

eshokrgozar:
When the pick_environment hook returns None, this along with [this PR in tk-core](https://github.com/shotgunsoftware/tk-core/pull/799) will allow us to only show a warning instead of the entire stack trace every time someone navigates to the unsupported entity page in SG. 

The reason behind all of this is related to [this other PR](https://github.com/shotgunsoftware/tk-framework-desktopserver/pull/77)

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -10,7 +10,7 @@
 
 # Metadata defining the behavior and requirements for this framework
 
-# expected fields in the configuration file for this engine
+# Expected fields in the configuration file for this engine
 configuration:
     browser_integration_hook:
         type: hook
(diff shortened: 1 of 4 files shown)


thebeeland:
A note about the MacOS and Windows failures before I merge: The Python 3.x CI on MacOS looks like it's failing to find the Python executable. For the Python 2.x failures, we're getting OpenSSL issues. Neither look like a result of problems with desktopserver, so I'm merging this anyway. We'll track down the CI issues separately.

PR #80, opened 2021-01-21
SG-19972 Fix for where we would continue to load actions when the list was empty.

pscadding:
If an instance of the tk-shotgun engine was configured with no apps, or the apps didn't show in this particular instance, then it would cause an infinite loop, because we treated an empty list as if we needed to recache the actions.


eshokrgozar (approved):
Makes sense. We might need to add this case to our testing since it was never caught. I'm not sure we do any automated testing on it but maybe we can at least flag it with Francis.

PR #84, opened 2021-09-16
SG-21431 update autobahn and cryptography

000paradox000:
Goals of this PR:

- Update some third-party dependencies where we can highlight
  cryptography, autobahn and certifi.

- Review and improve the process for updating the dependencies.

- Refresh the certificates files using as fixtures in unittesting.

Steps followed:

- Add recent versions of python 3.7 for azure pipeline
- Add linux binaries for python 2.7 and 3.7
- Add windows binaries for python 2.7 and 3.7
- Add mac binaries for python 2.7 and 3.7
- Replace build folder with newer packages and layout
- Fix recreation of bin folder
- Fix windows and linux binary installation scripts
- Add sources packages for python 2.7 and 3.7


jfboismenu:
This was awesome. I could not review the files on Github because they were too many so I had to do this:
`git diff --name-only origin/master | grep -v "python/bin" | grep -v "python/source" | grep -v "python/build" | xargs -n1 git diff origin/master —`

Could you python/source to python/src? We’ve added a folder under it and I’m worried we might open ourselves to path limit problems on Windows again.

`src/2.7/` is almost the same length as `source/`

PS: I can't approve the branch via the Add Review button because the tab locks up due to too many files, so take this comment as a "Request Changes".

jfboismenu:
Thanks, I approve! @mathurf , take it away! 😃

PR #135, opened 2023-07-06
SG-31325 Replace cgi module with the html module to support the escape method.

NorberMV:
This PR fix the AttributeError: module 'cgi' has no attribute 'escape', due to the escape method is no longer supported by the cgi module. Also included  Python 3.7.17 to the azure python versions list required by the macOS image in the Azure pipeline.


carlos-villavicencio-adsk (approved):
LGTM!

PR #136, opened 2023-07-11
SG-31201  Update requirements scripts

carlos-villavicencio-adsk:
Automates the source and binary dependency update from linux, mac (intel), and windows machines.

1. Just update the 3.7 and 3.9 requirements.txt file as [documented](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/master/resources/python/README.md#in-mac-update-requirementstxt-files).
2. Open a PR
3. Open a new PR with the newly created branch (with the `-automated` sufix) (CI uses a PAT I've created).


tests/integration_tests/interpreters.py
```diff
@@ -56,13 +56,15 @@ def get_python_interpreter_by_major_version(cls, major):
             "3.7.13",
             "3.7.14",
             "3.7.15",
+            "3.7.17",
```
julien-lang, line 59: This list keeps growing. Can we cleanup the old ones? Like the two oldest ones for instance? 
Like what are the criteria?

resources/python/pipelines/pipelines.yml
```diff
@@ -0,0 +1,179 @@
...
+  condition: not(endsWith( variables['System.PullRequest.SourceBranch'], '-automated'))
+  pool:
+    vmImage: 'macOS-12'
+  steps:
+  - template: template.yml
+    parameters:
+      python_version: 3.7
+  - script: |
+      git checkout $(System.PullRequest.SourceBranch)
+      git checkout -b ${{ parameters.branch }}
```
julien-lang, line 22: You can combine these 2 lines into one:
```yml
git checkout -b ${{ parameters.branch }} $(System.PullRequest.SourceBranch)
```

You might also want to use `git switch` instead if these VMs are using a recent enough version of GIT.

Not important or mandatory at all ;)

resources/python/install_binary_linux.sh
```diff
@@ -22,7 +22,7 @@ rm -rf $bin_dir
 mkdir $bin_dir
 
 # Install packages
-python build/pip install --target $bin_dir --no-deps -r $requirements
```
julien-lang, line 25: Why did we originally have pip used in the build folder?

Also, can you make these scripts run in "set -e" please? (and equivalent for PowerShell) Because we want to make sure they stop at the first error since we are automating everything.

resources/python/pipelines/pipelines.yml
```diff
@@ -0,0 +1,179 @@
...
+  steps:
+  - template: template.yml
+    parameters:
+      python_version: 3.9
+      branch: ${{ parameters.branch }}
+  - script: |
+      git checkout ${{ parameters.branch }}
+      python update_requirements.py
+      ./install_source_only.sh
+      git pull
```
julien-lang, line 45: Why do you pull here? If we are not synced, should fetch earlier instead?

resources/python/pipelines/pipelines.yml
```diff
@@ -0,0 +1,179 @@
...
+    workingDirectory: resources/python
+  - powershell: .\install_binary_windows.ps1
+    displayName: Run PowerShell Scripts
+    workingDirectory: resources/python
+  - script: |
+      git pull
+      git commit -am "Update binary requirements in Windows Python 3.9"
+      git push origin ${{ parameters.branch }}
+    displayName: Run Push Scripts
+    workingDirectory: resources/python
```
julien-lang, line 179: Looks like a missing empty line at EOF.

julien-lang (commented):
Looks good. Just a few comments.

resources/python/install_binary_linux.sh
carlos-villavicencio-adsk: I'm not familiar with the older syntax (`python build/pip...`), but it seems it was implemented some years ago.

resources/python/pipelines/pipelines.yml
carlos-villavicencio-adsk: I think I'm going to remove that line. When I was implementing this I had some race conditions, but now the jobs are dependent from each other.

resources/python/pipelines/pipelines.yml
carlos-villavicencio-adsk: Good call!

tests/integration_tests/interpreters.py
julien-lang: Works for me. Can you create a ticket about that, please?

julien-lang (approved):
LGTM!

PR #157, opened 2023-08-17
SG-31202 Address setuptools CVE

carlos-villavicencio-adsk:
Remove `build` directory that had fixed versions for pip, setuptools, and wheel. We're leaving that to the Python environment from now.

This was only used for installing the script for the source requirements.


julien-lang (approved):
LGTM!

PR #186, opened 2023-11-27
SG-33714 Upgrade Twisted to 23.10.0 for Python 3.9 only

carlos-villavicencio-adsk:
Twisted v23.10.0 stops support for Python 3.7. So we're keeping v22.10.0 for that specific version. We're also keeping that version for CI otherwise it breaks while installing on a Python 3.7 runner)

Resolves #178 
Resolves #184 
Resolves #185

Also, while testing this on CI I found some deprecation warnings on Python 3.10
Screenshot 2023-11-27 at 11 02 34


julien-lang (approved):
LGTM!

PR #181, opened 2023-11-23
SG-33306 bump cryptography automated

carlos-villavicencio-adsk:
Resolves #166
Resolves #167 
Resolves #168 
Resolves #169 
Resolves #170
Resolves #173 
Resolves #174 
Resolves #176
Resolves #177

PR #187, opened 2023-11-27
SG-33714 Upgrade twisted py39 automated

carlos-villavicencio-adsk:
Generated from #186


julien-lang (commented):
Why do I see changes in cryptography and autobhan while the explicit requirement file did not update them?

carlos-villavicencio-adsk:
@julien-lang that's weird to me too. Those changes should have been present on #181 😕

carlos-villavicencio-adsk:
On second thought, perhaps I generated #181 when upgrading cryptography 41.0.4 and then rebased the base branch with 41.0.5 but the latter didn't take effect. 
Anyway, this PR syncs everything.

PR #189, opened 2023-11-27
Fix 310 deprecation automated

carlos-villavicencio-adsk:
Installs Cryptography 41.0.5 from #181


python/tk_framework_desktopserver/command.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, p_out, target_queue):
         :param p_out: Pipe to read.
         :param target_queue: Queue that will accumulate the pipe output.
         """
-        Thread.__init__(self)
```
julien-lang, line 38: What's that? It's off-topic right?

julien-lang (changes requested):

python/tk_framework_desktopserver/command.py
carlos-villavicencio-adsk: I changed the base branch we can have an accurate diff.

julien-lang (approved):
LGTM!

PR #198, opened 2024-01-19
SG-34279 use manylinux 2.17 whl

carlos-villavicencio-adsk:
tk-framework-desktopserver v1.6.2 updated [Cryptography 41.0.3](https://pypi.org/project/cryptography/#files) module which uses [cp37-abi3-manylinux_2_28_x86_64](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/v1.6.2/resources/python/bin/3.9/linux/cryptography-41.0.3.dist-info/WHEEL) WHEEL for both python 3.7 and 3.9 (this requires glibc 2.28 and breaks desktop menu items on CentOS7).

This PR attempts to fix this by explicitly installing a manylinux 2.17 compatible wheel on linux platforms.

### DIFF
--- a/resources/python/update_requirements.py
+++ b/resources/python/update_requirements.py
@@ -217,20 +217,19 @@ class Updater(object):
             for dependency in dependencies:
                 package_name = dependency.split("==")[0]
 
+                # Cryptography is a special use case for Linux.
+                # Prioritized wheels require GLIBC 2.28 (manylinux_2_28_x86_64) which is not available on CentOS 7.
+                # Then, we download manylinux_2_17_x86_64 from PyPI.
+                if package_name == "cryptography":
+                    dependency = "cryptography==41.0.7 ; sys_platform != 'linux'\n"
+                    dependency += "https://files.pythonhosted.org/packages/14/fd/dd5bd6ab0d12476ebca579cbfd48d31bd90fa28fa257b209df585dcf62a0/cryptography-41.0.7-cp37-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl ; sys_platform == 'linux'"
+
+
                 # Figure which type of dependency it is and write
                 # it to the right requirements file.
                 requirement_to_add = [dependency + "\n"]
                 if package_name in self._binary_distributions:
-                    if package_name == "cryptography":
-                        # Cryptography is a special use case for linux.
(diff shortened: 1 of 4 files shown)


resources/python/update_requirements.py
```diff
@@ -221,7 +221,16 @@ def _update_requirements_file(self, dependencies):
...
                 requirement_to_add = [dependency + "\n"]
                 if package_name in self._binary_distributions:
-                    bin_handler.writelines(requirement_to_add)
+                    if package_name == "cryptography":
+                        # Cryptography is a special use case for linux.
+                        # Default wheels requires GLIBC 2.28 (manylinux_2_28_x86_64) which is not available on CentOS 7.
+                        cryptography_to_add = [
+                             "cryptography==41.0.7 ; sys_platform != 'linux'\n",
+                             "https://files.pythonhosted.org/packages/14/fd/dd5bd6ab0d12476ebca579cbfd48d31bd90fa28fa257b209df585dcf62a0/cryptography-41.0.7-cp37-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl ; sys_platform == 'linux'\n"
+                        ]
```
julien-lang, line 230: Looks good but could you move this test above the "# Figure which type of dependency it is and write" comment to make the code easier to read (in my opinion):

```python
for dependency in dependencies:
    package_name = dependency.split("==")[0]

    if package_name == "cryptography":
        # Cryptography is a special use case for linux.
        # Default wheels requires GLIBC 2.28 (manylinux_2_28_x86_64) which is not available on CentOS 7.
        dependency = "cryptography==41.0.7 ; sys_platform != 'linux'\n"
        dependency+= "https://files.pythonhosted.org/packages/14/fd/dd5bd6ab0d12476ebca579cbfd48d31bd90fa28fa257b209df585dcf62a0/cryptography-41.0.7-cp37-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl ; sys_platform == 'linux'"

    # Figure which type of dependency it is and write
    # it to the right requirements file.
    requirement_to_add = [dependency + "\n"]
    if package_name in self._binary_distributions:

...
```

I feel that, this way, the 2 different behaviors (whether the package is crypto, and whether the package is src or binary) are more separated and the overall function is easier to read.
I might be wrong.

julien-lang (approved):
LGTM!

PR #214, opened 2024-02-20
SG-34580 Add Python 3.10 sources and binaries

carlos-villavicencio-adsk:


resources/python/README.md
```diff
@@ -8,14 +8,23 @@ Officially Supported Python Versions:
 - Mac 
   - 3.7.16
   - 3.9.16
+  - 3.10.13
 - Windows 
   - 3.7.9
   - 3.9.13
+  - 3.10.11
```
julien-lang, line 15: Stupid question: why don't we use the same minor version for Windows?

resources/python/README.md
```diff
@@ -211,7 +182,7 @@ git push
 
 ### In every Operating System, activate virtualenvs and execute the corresponding script to install binaries  and then push changes to repository
 
-Mac
+#### Mac
```
julien-lang, line 185: Minor

```suggestion
#### macOS
```

resources/python/pipelines/pipelines.yml
```diff
@@ -50,11 +50,31 @@ jobs:
     displayName: Generate all explicit_requirements & Update source requirements
     workingDirectory: resources/python
 
+- job: install_source_dependencies_3_10
+  displayName: Install source dependencies Python 3.10
```
julien-lang, line 54: Minor: could you move this block up above Py39 to keep the same DESC order please?
This should also simplify the changes of "dependsOn"

tests/integration_tests/interpreters.py
julien-lang, line 1: Thank you! :dancers:

julien-lang (commented):
A few minor comments. Otherwise, looks good.

resources/python/README.md
carlos-villavicencio-adsk: It is not available for Windows: https://www.python.org/downloads/windows/

resources/python/pipelines/pipelines.yml
carlos-villavicencio-adsk: I think doing so also will need to change the order between 3.7 and 3.9 we currently had.

I understand your POV. However, I think the blocks could be in the actual execution order to ease the debugging. Do you think it makes sense?

julien-lang: It does not matter. It would just have helped the review process. I'll let you decide what you think is best.

carlos-villavicencio-adsk: Let's keep the execution order for now until we simplify this a bit.

julien-lang (approved):
LGTM!

julien-lang (approved):

PR #235, opened 2024-05-09
Improve error handling

carlos-villavicencio-adsk:
Sometimes `reason.value` doesn't have `message` attribute, raising a different error. This quick PR attempts to fix this.

### DIFF
--- a/python/tk_framework_desktopserver/server_protocol.py
+++ b/python/tk_framework_desktopserver/server_protocol.py
@@ -89,7 +89,7 @@ class ServerProtocol(WebSocketServerProtocol):
             # Known certificate error. These work for firefox and safari, but chrome rejected certificate
             # are currently indistinguishable from lost connection. This is true as of July 21st, 2015.
             certificate_error = False
-            if hasattr(reason.value, "message"):
+            try:
                 certificate_error |= (
                     reason.type is OpenSSL.SSL.Error
                     and reason.value.message[0][2] == "ssl handshake failure"
@@ -98,6 +98,8 @@ class ServerProtocol(WebSocketServerProtocol):
                     reason.type is OpenSSL.SSL.Error
                     and reason.value.message[0][2] == "tlsv1 alert unknown ca"
                 )
+            except AttributeError:
+                logger.warning("Unexpected error message in object: {}".format(reason))
 
             certificate_error |= bool(reason.check(error.CertificateError))
             certificate_error |= bool(reason.check(error.ConnectionLost))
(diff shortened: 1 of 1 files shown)


python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -89,15 +89,18 @@ def connectionLost(self, reason):
...
             certificate_error = False
-            certificate_error |= (
-                reason.type is OpenSSL.SSL.Error
-                and reason.value.message[0][2] == "ssl handshake failure"
-            )
-            certificate_error |= (
-                reason.type is OpenSSL.SSL.Error
-                and reason.value.message[0][2] == "tlsv1 alert unknown ca"
-            )
+            if hasattr(reason.value, "message"):
```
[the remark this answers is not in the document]
julien-lang: Yes, I kind of agree here. We should try to retrieve "reason.value.message[0][2]" in a try/except block and then run the following lines. It will be even more fault tolerant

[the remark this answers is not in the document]
carlos-villavicencio-adsk: Good call guys, thank you

python/tk_framework_desktopserver/server_protocol.py
```diff
@@ -89,15 +89,20 @@ def connectionLost(self, reason):
...
-            certificate_error |= (
-                reason.type is OpenSSL.SSL.Error
-                and reason.value.message[0][2] == "ssl handshake failure"
-            )
-            certificate_error |= (
-                reason.type is OpenSSL.SSL.Error
-                and reason.value.message[0][2] == "tlsv1 alert unknown ca"
-            )
+            try:
+                certificate_error |= (
```
eduardoChaucaGallegos, line 93: Is it possible to use only one line to get message? :
`message = reason.value.message[0][2]`
and then use "message" in comparisons?

julien-lang: I agree, could you extract the message in a small try/except block and then use it:

```python
try:
  message = reason.value.message[0][2]
except AttributeError, IndexError:
  logger.warning("Unexpected error message in object: {}".format(reason))
else:
  certificate_error |= (
    reason.type is OpenSSL.SSL.Error
    and message == "ssl handshake failure"
  )
  certificate_error |= (
    reason.type is OpenSSL.SSL.Error
    and message == "tlsv1 alert unknown ca"
  )
```

carlos-villavicencio-adsk: Sounds good. My intention wasn't to do a full refactor but I can take the chance.

eduardoChaucaGallegos (approved):

julien-lang (approved):
LGTM!

However, I wonder about this function's behavior and why having a big try/except capturing everything.

PR #257, opened 2024-08-15
SG-36027 Bundle modules for Python 3.11

carlos-villavicencio-adsk:
- Refactor some stuff to make it easier to add new versions in the future
- Print messages in console when a version is not available. Useful for debugging.
- Downgrade `setuptools` for Python 3.7 that was causing issues.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,14 +5,18 @@ ignore:
     - resources/python/bin/3.7/explicit_requirements.txt
     - resources/python/bin/3.9/explicit_requirements.txt
     - resources/python/bin/3.10/explicit_requirements.txt
+    - resources/python/bin/3.11/explicit_requirements.txt
     - resources/python/src/3.7/explicit_requirements.txt
     - resources/python/src/3.9/explicit_requirements.txt
     - resources/python/src/3.10/explicit_requirements.txt
+    - resources/python/src/3.11/explicit_requirements.txt
 exclude:
   global:
     - resources/python/bin/3.7/explicit_requirements.txt
     - resources/python/bin/3.9/explicit_requirements.txt
     - resources/python/bin/3.10/explicit_requirements.txt
+    - resources/python/bin/3.11/explicit_requirements.txt
     - resources/python/src/3.7/explicit_requirements.txt
     - resources/python/src/3.9/explicit_requirements.txt
     - resources/python/src/3.10/explicit_requirements.txt
+    - resources/python/src/3.11/explicit_requirements.txt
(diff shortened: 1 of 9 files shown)


.snyk
```diff
@@ -5,14 +5,18 @@ ignore:
     - resources/python/bin/3.7/explicit_requirements.txt
     - resources/python/bin/3.9/explicit_requirements.txt
     - resources/python/bin/3.10/explicit_requirements.txt
+    - resources/python/bin/3.11/explicit_requirements.txt
     - resources/python/src/3.7/explicit_requirements.txt
     - resources/python/src/3.9/explicit_requirements.txt
     - resources/python/src/3.10/explicit_requirements.txt
+    - resources/python/src/3.11/explicit_requirements.txt
```
julien-lang, line 12: No wildcards possible?

python/tk_framework_desktopserver/__init__.py
```diff
@@ -25,13 +26,24 @@
...
-    sys.path.insert(0, os.path.join(binaries_path, _version_dir, "mac"))
+    binaries_os_path = os.path.join(binaries_path, _version_dir, "mac")
 elif sgtk.util.is_windows():
-    sys.path.insert(0, os.path.join(binaries_path, _version_dir, "win"))
+    binaries_os_path = os.path.join(binaries_path, _version_dir, "win")
 elif sgtk.util.is_linux():
-    sys.path.insert(0, os.path.join(binaries_path, _version_dir, "linux"))
+    binaries_os_path = os.path.join(binaries_path, _version_dir, "linux")
+
+if os.path.exists(binaries_os_path):
```
julien-lang, line 35: This variable might not be defined when used here given the condition above.

Not super important but I understand that this PR is about Fault Tolerance to make sure this does not crash.

Also, when we don't find the path, should we "exit" or "raise" instead of continuing ignoring the error. Because it will just crash later isn't it?

resources/python/update_requirements.py
julien-lang, line 1: I like this change!

python/tk_framework_desktopserver/__init__.py
carlos-villavicencio-adsk: It should only be undefined if this is running in any other OS than macos, linux, and windows. But you're right a good practice is to define it at least with `None` value.

I wasn't sure about `raise` here on error. We're sure it will crash later. I like this idea, will implement it.

.snyk
carlos-villavicencio-adsk: Unfortunately nothing in [their documentation](https://docs.snyk.io/manage-risk/policies/the-.snyk-file).

julien-lang (changes requested):
Just waiting for the "raise" part. Everything else looks good.

carlos-villavicencio-adsk:
> Just waiting for the "raise" part. Everything else looks good.

It's already implemented in python/tk_framework_desktopserver/__init__.py

julien-lang (approved):
LGTM!

PR #266, opened 2024-10-24
SG-36672 Use zip packages instead PoC

carlos-villavicencio-adsk:
This PR only packages source dependencies. Binary dependencies are not taken into account because Python disallows dynamic extensions inside a ZIP file per [PEP 273](https://peps.python.org/pep-0273/#subdirectory-equivalence). [Reference](https://stackoverflow.com/questions/44248393/why-is-the-import-of-so-files-from-zip-files-disallowed-in-python).

This PR changes the automation scripts. The actual removal is now in #268

### DIFF
--- a/resources/python/install_source_only.sh
+++ b/resources/python/install_source_only.sh
@@ -21,7 +21,8 @@ echo "----------------------------------------------------"
 echo "Set base paths"
 
 requirements_filename="explicit_requirements.txt"
-build_dir=$PWD/build
+package_filename="pkgs.zip"
+build_dir=$PWD
 source_dir="src/$python_version"
 source_requirements="$source_dir/$requirements_filename"
 
@@ -31,7 +32,7 @@ echo "Source Requirements: $source_requirements"
 echo "----------------------------------------------------"
 echo "Remove current packages"
 
-find $source_dir/* ! -name $requirements_filename -maxdepth 1 -exec rm -rf {} +
+find $source_dir/* ! -name $package_filename ! -name $requirements_filename -maxdepth 1 -exec rm -rf {} +
 
 echo "----------------------------------------------------"
 echo "Install new packages"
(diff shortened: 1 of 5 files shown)


resources/python/install_source_only.sh
```diff
@@ -59,6 +60,14 @@ rm -Rf $source_dir/incremental/tests
 # into other processes, which we don't require. That module is compiled, so we'll delete it.
 rm -Rf $source_dir/twisted/python/_sendmsg.so
 
+# Compress all files
+cd $source_dir
+zip -q -r pkgs.zip ./*
+cd $build_dir
```
julien-lang, line 66: I prefer the pushd/popd approach. Or I also recently learned `cd -`

resources/python/install_source_only.sh
```diff
@@ -59,6 +60,14 @@ rm -Rf $source_dir/incremental/tests
...
+
+# Remove files
+find $source_dir/* ! -name $package_filename ! -name $requirements_filename -maxdepth 1 -exec rm -rf {} +
```
julien-lang, line 69: I wonder if that would do the job:

```suggestion
find $source_dir/* ! -name $package_filename ! -name $requirements_filename -delete
```

julien-lang (commented):
Looks good. JUst a few minor comments.

resources/python/install_source_only.sh
carlos-villavicencio-adsk: that leaves empty directories all around. Not useful.

resources/python/install_source_only.sh
carlos-villavicencio-adsk: Can you elaborate more?

julien-lang:
```bash
pushd $source_dir
zip -q -r pkgs.zip ./*
popd
```

carlos-villavicencio-adsk: TIL pushd/popd. Thanks!

resources/python/pipelines/pipelines.yml
julien-lang, line 1: We need a different PR for that I think.

carlos-villavicencio-adsk: Ok let's rollback

julien-lang (approved):

carlos-villavicencio-adsk:
Closed because changes were also included in #268

PR #270, opened 2024-11-29
SG-36829 SG-36830 - Upgrade Twisted to 24.10.0

carlos-villavicencio-adsk:
Same as last year #190.

Standardize the .snyk file to match the projects on Snyk. Update to exclude directories that belong to Python 3.7 since it's no longer supported.

PR #271, opened 2024-11-29
SG-36830 Upgrade Twisted -  automated

carlos-villavicencio-adsk:
Generated from #270


resources/python/src/3.7/pkgs.zip
julien-lang, line 1: Why do we remove this file?

resources/python/bin/3.11/linux/zope/interface/_zope_interface_coptimizations.cpython-311-x86_64-linux-gnu.so
julien-lang, line 1: Why do we remove this file?

Also what about Windows and macOS?

carlos-villavicencio-adsk: I don't think this was removed. Just updated.

resources/python/src/3.7/pkgs.zip
carlos-villavicencio-adsk: I don't think this was removed. Just updated.

julien-lang: Yes, indeed. Sorry about that!

resources/python/bin/3.11/linux/zope/interface/_zope_interface_coptimizations.cpython-311-x86_64-linux-gnu.so
julien-lang: Ok. What about mac and Win?

julien-lang: The fact that the linux libs was updated but not the Windows or macOS one is suspicious. So I'm raising a question: is this expected or is it possible there is a glitch in the process?

carlos-villavicencio-adsk: It seems to work as expected on Linux Rocky 8. 
image

julien-lang: Sorry if I am not clear. Trying to 🤔!

What I mean is: why don't we have the Windows/macOS libs updated too? But only the Linux one? That does not make sense, does it?

carlos-villavicencio-adsk: I don't really know.

julien-lang (approved):
LGTM!

PR #274, opened 2025-01-08
SG-37774 Remove Python 2 syntax

carlos-villavicencio-adsk:
- Remove specific Python 2 syntax
- Remove tests dependencies


tests/integration_tests/interpreters.py
```diff
@@ -172,15 +164,8 @@ def setUpClass(cls):
...
-        if sys.version_info.major == 2:
-            cls.python_config = cls.create_pipeline_config_for_python_version(
-                "python2", 2
-            )
-        if sys.version_info.major == 3:
-            cls.python_config = cls.create_pipeline_config_for_python_version(
-                "python3", 3
-            )
+        # Create a config for python3
+        cls.python_config = cls.create_pipeline_config_for_python_version("python3", 3)
```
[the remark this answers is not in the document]
carlos-villavicencio-adsk: I changed this and it worked. No more `python3` reference. Good call.

julien-lang (approved):

python/tk_framework_desktopserver/command.py
```diff
@@ -231,10 +231,10 @@ def _call_cmd_win32(args, env):
 
             # Read back the output from the two.
             with open(stdout_path, "rt") as stdout_file:
-                stdout_lines = [sgutils.ensure_str(l) for l in stdout_file]
+                stdout_lines = [l for l in stdout_file]
 
             with open(stderr_path) as stderr_file:
-                stderr_lines = [sgutils.ensure_str(l) for l in stderr_file]
+                stderr_lines = [l for l in stderr_file]
```
eduardoChaucaGallegos, line 237: if we are not going to use `sgutils `then we can delete the import block
```
try:
    from tank_vendor import sgutils
except ImportError:
    from tank_vendor import six as sgutils
```

carlos-villavicencio-adsk: We still use it on the `ReadThread::run` method

```py
while True:
            line = sgutils.ensure_str(self.pipe.readline())  # <-----
            if line == "":
                break
            self.target_queue.put(line)
```

However, I'm not sure to replace it with `str`. Do you have experience with standard out pipes return type for `readline`?

eduardoChaucaGallegos (approved):

PR #294, opened 2025-04-08
Add PR template

carlos-villavicencio-adsk:


julien-lang (commented):
Is it possible to have a template for the whole organization? 
Or for a set list of repositories?

julien-lang:
> @julien-lang it seems [it's possible](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file). However, going this path we cannot make exceptions and will affect every repo. Although every repo can override its contents.

Yes, I think I like this option.

PR #299, opened 2025-09-04
SG-39655 Update Python third party libraries (automated)

julien-lang:
Automated CI process triggered by #298 fixing a GLIBC bug for EL8.

Here is the biggest CI change in log that we were looking for:
```diff
--- before.log	2025-09-03 13:58:55.313132692 -0700
+++ after.log	2025-09-03 13:59:08.835220943 -0700
@@ -1,6 +1,6 @@
 Collecting cffi==1.15.1
   Using cached cffi-1.15.1-cp37-cp37m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (427 kB)
 Collecting cryptography==44.0.1
-  Using cached cryptography-44.0.1-cp37-abi3-manylinux_2_34_x86_64.whl (4.2 MB)
+  Using cached cryptography-44.0.1-cp37-abi3-manylinux_2_28_x86_64.whl (4.2 MB)
 Collecting zope.interface==5.5.2
   Using cached zope.interface-5.5.2-cp37-cp37m-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_12_x86_64.manylinux2010_x86_64.whl (254 kB)
```

Of course, there are also a few more minor version jumps.


carlos-villavicencio-adsk:
Don't forget the important notice for Apple Silicon [here](https://github.com/shotgunsoftware/tk-framework-desktopserver/blob/master/resources/python/README.md#macos).

Azure CI is based on Intel architecture. Because of that, some shared objects (*.so files) are fetched according to that. **This will break on Apple Silicon users.**

Unfortunately, I can't envision how your solution can also help in this case because these third parties don't want to provide a fat binary (combined architecture). That's what the readme is instructing how we can do it ourselves. I can share these binaries internally with you if you want. I always do this manual step on these automated branches. Probably it can be automated once we have a native Apple Silicon worker.

PR #360, opened 2026-07-22
SG-30953 Remove Python 3.7 support

stevelittlefish:
## Summary
- Delete `resources/python/{src,bin}/3.7/` - the runtime loader in `tk_framework_desktopserver/__init__.py` builds its path from `sys.version_info` and raises a clear `RuntimeError` if the version dir is missing, so Python 3.7 now fails explicitly instead of silently running unsupported code.
- Remove the 4 Python 3.7 jobs from the per-PR package-regeneration pipeline (`resources/python/pipelines/pipelines.yml`): source deps, and Mac/Linux/Windows binary deps. Re-pointed the `dependsOn` chain for each group's new head job, and moved the one-time "create automated branch" step from the old 3.7 job into the new head job (3.9) since nothing else in the chain creates that branch.
- Simplify `resources/python/requirements.txt` and `azure-pipelines/requirements.txt`: removed now-dead `python_version < "3.9"` branches (pyOpenSSL, Twisted, cffi, cryptography, pyasn1, websocket-client) and their now-always-true `python_version >= "3.9"` counterparts.
- Update `resources/python/README.md`: drop 3.7 from the supported version list, shift the manual walkthrough's example version from 3.7 to 3.9.

## Verification
- No remaining Python 3.7 references outside of unrelated third-party code (PyPy/CPython compat comments, OID numbers) inside the still-bundled 3.9/3.10/3.11/3.13 packages.
- Validated the regeneration pipeline's job graph with a YAML parse after edits: 16 jobs, no dangling `dependsOn` references, no leftover `3_7` job names.
- Confirmed the runtime loader still resolves `resources/python/{bin,src}/3.9/...` correctly.
- Full test suite requires the `sgtk`/`tank_test` toolkit harness, not available standalone here - recommend running in CI before merge.

Part of epic SG-36195.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


carlos-villavicencio-adsk (approved):
LGTM

